PR #76, opened 2020-04-07
Fix move constructor implementation

mathbagu:

N/A

Bug fix

The move constructor of a network doesn't update the back references to all identifiables

The back reference is updated at the end of the move constructor

**Other information**:

### DIFF
--- a/include/powsybl/iidm/Identifiable.hpp
+++ b/include/powsybl/iidm/Identifiable.hpp
@@ -25,11 +25,15 @@ public: // Validable
     std::string getMessageHeader() const override;
 
 public:
-    Identifiable(const Identifiable& identifiable) = delete;
+    Identifiable(const Identifiable&) = delete;
+
+    Identifiable(Identifiable&& identifiable) noexcept;
 
     ~Identifiable() noexcept override = default;
 
-    Identifiable& operator=(const Identifiable& identifiable) = delete;
+    Identifiable& operator=(const Identifiable&) = delete;
+
+    Identifiable& operator=(Identifiable&&) noexcept = delete;
 
     const std::string& getId() const;
 
@@ -48,8 +52,6 @@ public:
     stdcxx::optional<std::string> setProperty(const std::string& key, const std::string& value);
 
 protected:
-    Identifiable(Identifiable&&) = default;
-
     Identifiable(const std::string& id, const std::string& name);
 
 private:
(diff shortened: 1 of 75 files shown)


include/powsybl/iidm/Connectable.hpp
```diff
@@ -24,8 +24,16 @@ class Network;
 
 class Connectable : public Identifiable, public MultiVariantObject {
 public:
+    Connectable(const Connectable& connectable) = delete;
+
+    Connectable(Connectable&& connectable) noexcept;
```
sebalaig, line 29: I'm not sure we have to provide this body and let the compiler generate it

include/powsybl/iidm/Identifiable.hpp
```diff
@@ -25,11 +25,15 @@ class Identifiable : public virtual Validable, public Extendable {
     std::string getMessageHeader() const override;
 
 public:
-    Identifiable(const Identifiable& identifiable) = delete;
+    Identifiable(const Identifiable&) = delete;
+
+    Identifiable(Identifiable&&) noexcept;
```
mathbagu, line 30: clang-tidy

include/powsybl/iidm/Network.hpp
```diff
@@ -79,10 +80,16 @@ class Network : public Container, public VariantManagerHolder {
 public:
     Network(const std::string& id, const std::string& sourceFormat);
 
-    Network(Network&&) = default;
+    Network(const Network&) = delete;
+
+    Network(Network&&) noexcept;
```
mathbagu, line 85: clang-tidy

include/powsybl/iidm/Connectable.hpp
mathbagu: If I remember well, you cannot use a default implementation because you inherit from virtual base classes.

include/powsybl/iidm/VariantManager.hpp
```diff
@@ -30,13 +30,17 @@ class VariantManager {
 public:
     explicit VariantManager(Network& network);
 
-    VariantManager(const VariantManager& variantManager) = delete;
+    VariantManager(const VariantManager&) = delete;
```
sebalaig, line 33: I don't know if it is clang-tidy compliant (every parameter should be named), I suppose it is. In such case, I think we should be consistent and also remove the parameter of the move constructor.

Issue #100, opened 2020-05-12
Add support for import/export properties


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
To customize the import/export, we have to create an `ImportOptions` or and `ExportOptions` objects. But in Java, the importer is generic and is able to load the options from a property map. It could be nice to have the same kind of feature.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
Be able to set the options using properties instead of hard code things.

* **What is the motivation / use case for changing the behavior?**
Allow the user program to load the configuration from a file instead of hard code every options.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

Issue #43, opened 2020-03-11
[IIDM v1.0] Support all XML export options


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
There is no equivalent to the java XMLImporter/XMLExporter parameters.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
The XML import/export should be customizable, with exactly the same feature than in the Java version.

* **What is the motivation / use case for changing the behavior?**
Have both Java and C++ API as close as possible

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #106, opened 2020-05-18
Xml import/export options from a Properties map

sebalaig:

Issue #100 and #43

This PR add the ability to import/export a network from a map of options

User has to set options by filling ImportOptions/ExportOptions

Options can be set by writing something like this : 
```
    converter::Properties properties;
    properties.put(THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND, "true");
    properties.put(TOPOLOGY_LEVEL, "BUS_BREAKER");

    FakeAnonymizer anonymizer;
    const Network& network = Network::readXml(stream, properties, anonymizer);
```

**Other information**:


include/powsybl/iidm/converter/Parameter.hpp
```diff
@@ -0,0 +1,89 @@
...
+namespace converter {
+
+static const char* const ANONYMISED = "iidm.export.xml.anonymised";
+static const char* const EXTENSIONS_LIST = "iidm.export.xml.extensions";
+static const char* const INDENT = "iidm.export.xml.indent";
+static const char* const ONLY_MAIN_CC = "iidm.export.xml.only-main-cc";
+static const char* const THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND = "iidm.export.xml.throw-exception-if-extension-not-found";
+static const char* const TOPOLOGY_LEVEL = "iidm.export.xml.topology-level";
+static const char* const VERSION = "iidm.export.xml.version";
+static const char* const WITH_BRANCH_STATE_VARIABLES = "iidm.export.xml.with-branch-state-variables";
```
mathbagu, line 30: This constants are specific to the XIIDM import/exporter so you have to move them in the `iidm.converter.xml` namespace

include/powsybl/iidm/converter/Parameter.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+enum class ParameterType : char {
+    BOOLEAN,
+    STRING,
+    STRING_LIST
+};
```
mathbagu, line 36: Could be converted as a inner class of the parameter class. What do you think

include/powsybl/iidm/converter/Parameter.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+class Parameter {
+public:
+    Parameter() = delete;
```
mathbagu, line 40: Not necessary: declaring constructor removes the default constructor.

include/powsybl/iidm/converter/Parameter.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+    Parameter(const std::string& name, const ParameterType& type, const std::string& description, const std::string& defaultValue);
+
+    Parameter(const std::string& name, const ParameterType& type, const std::string& description, bool defaultValue);
+
+    Parameter(const std::string& name, const ParameterType& type, const std::string& description, const std::set<std::string>& defaultValue);
```
mathbagu, line 46: Consider passing everything using rvalue

include/powsybl/iidm/converter/Parameter.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+    ~Parameter() = default;
+
+    Parameter(const Parameter&) = default;
+
+    Parameter(Parameter&&) = default;
+
+    Parameter& operator=(const Parameter&) = default;
+
+    Parameter& operator=(Parameter&&) = default;
```
mathbagu, line 56: Not sure we should allow this

include/powsybl/iidm/converter/Parameter.hpp
```diff
@@ -0,0 +1,89 @@
...
+    const ParameterType& getType() const;
+
+private:
+    std::vector<std::string> m_names;
+
+    ParameterType m_type;
+
+    std::string m_description;
+
+    boost::any m_defaultValue;
```
mathbagu, line 78: I'm not sure we should use boost::any. To be discussed

include/powsybl/iidm/converter/Properties.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <map>
+#include <string>
+
+namespace powsybl {
+
+namespace iidm {
+
+namespace converter {
+
+class Properties {
```
mathbagu, line 20: I think we should use the same properties class in IIDM and in the converter. Consider to have only one class, and move this class in `stdcxx` namespace

src/iidm/converter/Parameter.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <powsybl/iidm/converter/Parameter.hpp>
+
+namespace powsybl {
+
+namespace iidm {
+
+namespace converter {
+
+Parameter::Parameter(const std::string& name, const ParameterType& type, const std::string& description, const std::string& defaultValue) :
+    m_type(type), m_description(description), m_defaultValue(defaultValue) {
```
mathbagu, line 17: Coding style: one assignment per line. This remark applies for all the constructors

src/iidm/converter/Parameter.cpp
```diff
@@ -0,0 +1,56 @@
...
+    m_names.emplace_back(name);
+}
+
+Parameter::Parameter(const std::string& name, const ParameterType& type, const std::string& description, const std::set<std::string>& defaultValue) :
+    m_type(type), m_description(description), m_defaultValue(defaultValue) {
+    m_names.emplace_back(name);
+}
+
+Parameter& Parameter::addAdditionalNames(const std::string& additionalName) {
```
mathbagu, line 31: In java, the addAdditionalNames takes an array of string.
```
public Parameter addAdditionalNames(String... names) {
        Objects.requireNonNull(names);
        this.names.addAll(Arrays.asList(names));

        return this;
    }
```

src/iidm/Network.cpp
```diff
@@ -43,6 +48,15 @@ namespace powsybl {
...
 
+static const converter::Parameter& ANONYMISED_PARAMETER = converter::Parameter(converter::ANONYMISED, converter::ParameterType::BOOLEAN, "Anonymise exported network", false);
+static const converter::Parameter& INDENT_PARAMETER = converter::Parameter(converter::INDENT, converter::ParameterType::BOOLEAN, "Indent export output file", true);
+static const converter::Parameter& EXTENSIONS_LIST_PARAMETER = converter::Parameter(converter::EXTENSIONS_LIST, converter::ParameterType::STRING_LIST, "The list of exported extensions", std::set<std::string>());
+static const converter::Parameter& ONLY_MAIN_CC_PARAMETER = converter::Parameter(converter::ONLY_MAIN_CC, converter::ParameterType::BOOLEAN, "Export only main CC", false);
+static const converter::Parameter& THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER = converter::Parameter(converter::THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND, converter::ParameterType::BOOLEAN, "Throw exception if extension not found", false);
+static const converter::Parameter& TOPOLOGY_LEVEL_PARAMETER = converter::Parameter(converter::TOPOLOGY_LEVEL, converter::ParameterType::STRING, "Export network in this topology level", std::string("NODE_BREAKER"));
```
mathbagu, line 56: Not sure you need to explicitly call the conversion from `char*` to `std::string`

src/iidm/Network.cpp
```diff
@@ -43,6 +48,15 @@ namespace powsybl {
...
+static const converter::Parameter& VERSION_PARAMETER = converter::Parameter(converter::VERSION, converter::ParameterType::STRING, "IIDM-XML version in which files will be generated", converter::xml::IidmXmlVersion::CURRENT_IIDM_XML_VERSION().toString("."));
+static const converter::Parameter& WITH_BRANCH_STATE_VARIABLES_PARAMETER = converter::Parameter(converter::WITH_BRANCH_STATE_VARIABLES, converter::ParameterType::BOOLEAN, "Export network with branch state variables", true);
+
```
mathbagu, line 59: There is an issue in the dependencies of the namespaces... I didn't notice that but IIDM cannot depend on the converter as converter depends on IIDM. We have to refactor the code.

Maybe the new `read` functions should be moved to the NetworkXml class.

src/iidm/Network.cpp
```diff
@@ -56,6 +70,15 @@ Network Network::readXml(std::istream& istream, const converter::ImportOptions&
     return converter::xml::NetworkXml::read(istream, options, anonymizer);
 }
 
+Network Network::readXml(std::istream& istream, const converter::Properties& parameters, const converter::Anonymizer& anonymizer) {
+    converter::ImportOptions options;
+    converter::Parameter THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_IMPORT_PARAMETER(THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER);
+    THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_IMPORT_PARAMETER.addAdditionalNames("throwExceptionIfExtensionNotFound");
```
mathbagu, line 76: That's uggly: Why don't you add the additional names when you declare the parameter

include/powsybl/iidm/converter/Parameter.hpp
sebalaig: Agree!

include/powsybl/iidm/converter/Parameter.hpp
sebalaig: Done except for defaultValue=bool and ParameterType because clang-tidy says they are trivially copiable : 
```
Clang-Tidy: Std::move of the variable 'type' of the trivially-copyable type 'powsybl::iidm::converter::Parameter::ParameterType' has no effect; remove std::move()
```

include/powsybl/iidm/converter/Parameter.hpp
sebalaig: Exact: moved in a file `powsybl/iidm/converter/xml/Constants.hpp`

src/iidm/Network.cpp
sebalaig: I agree... But the additional name is added in XMLImporter only (not in XMLExporter). 
I wanted to be Java strictly consistent ;)

src/iidm/Network.cpp
sebalaig: Yes it is needed, unless there is an implicit conversion into bool, clang-tidy highlights it:
```
/mnt/e/Projet_SRO/rte/powsybl-iidm4cpp/src/iidm/Network.cpp:57:202: error: implicit conversion 'const char *' -> bool [readability-implicit-bool-conversion,-warnings-as-errors]
static const converter::Parameter& TOPOLOGY_LEVEL_PARAMETER = converter::Parameter(converter::xml::TOPOLOGY_LEVEL, converter::Parameter::ParameterType::STRING, "Export network in this topology level", "NODE_BREAKER");
```

src/iidm/Network.cpp
sebalaig: PR !111 fixes the problem but you said in Java both libraries will be merged in a next future, so we can postpone the fix. No?

include/powsybl/iidm/converter/Parameter.hpp
sebalaig: Do you prefer using a single string to store it? Or duplicating explicitly typed members?

src/iidm/Network.cpp
mathbagu: In java the static initializer is used to add the additional names during the construction of the constant. In your proposal your create a copy of the parameter each time readXml() is called:
```
    private static final Parameter THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER
            = new Parameter(THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND, ParameterType.BOOLEAN, "Throw exception if extension not found", Boolean.FALSE)
            .addAdditionalNames("throwExceptionIfExtensionNotFound");
```

include/powsybl/iidm/converter/Parameter.hpp
mathbagu: I think storing the value in a string is OK. it's easy to convert a string to a boolean.

src/iidm/Network.cpp
mathbagu: I need to review your PR once again to decide. The PR would be smaller when the PR about properties will be merged.

include/powsybl/iidm/converter/Parameter.hpp
sebalaig: Ok to use a string.

> I think you should also remove the generic method: if someone ask for an unsupported type, how do you catch it?

Linker would fail :/
Do you mean replacing `ConversionParameters::readParameter` by triplet `ConversionParameters::readBooleanParameter`, `ConversionParameters::readStringParameter` et `ConversionParameters::readStringListParameter` ?

Or using a triple condition `std::enable_if` on it instead?

include/powsybl/iidm/Network.hpp
```diff
@@ -66,10 +72,14 @@ class Network : public Container, public VariantManagerHolder {
 
     static Network readXml(std::istream& istream, const converter::ImportOptions& options, const converter::Anonymizer& anonymizer);
 
+    static Network readXml(std::istream& istream, const stdcxx::Properties& parameters, const converter::Anonymizer& anonymizer);
```
mathbagu, line 75: It could be nice to propose two overloads without the anonymizer, or to add a constructor to ImportOptions and ExportOptions that takes a Properties object.

The idea is to keep the API user-friendly

include/powsybl/iidm/converter/ConversionParameters.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace converter {
+
+class Parameter;
+
+class ConversionParameters {
+public:
+    static bool readBooleanParameter(const stdcxx::Properties& parameters, const Parameter& configuredParameter);
+
+    static std::set<std::string> readStringListParameter(const stdcxx::Properties& parameters, const Parameter& configuredParameter);
```
mathbagu, line 32: Sometimes the order is important: change the return type to a `std::list` or `std::vector`

include/powsybl/iidm/converter/Parameter.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+namespace powsybl {
+
+namespace iidm {
+
+namespace converter {
+
+class Parameter {
+public:
+    enum class ParameterType : char {
```
mathbagu, line 26: If you declare this enum inner the Parameter class, rename it to Type (Parameter::Type), or extract it to its own header

include/powsybl/iidm/converter/Parameter.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+    ~Parameter() = default;
+
+    Parameter(const Parameter&) = default;
+
+    Parameter(Parameter&&) noexcept = default;
+
+    Parameter& operator=(const Parameter&) = default;
+
+    Parameter& operator=(Parameter&&) = delete;
```
mathbagu, line 47: Why do you delete this one? Maybe it's acceptable to declare it as default

include/powsybl/iidm/converter/Parameter.hpp
```diff
@@ -0,0 +1,81 @@
...
+    const ParameterType& getType() const;
+
+private:
+    std::vector<std::string> m_names;
+
+    ParameterType m_type;
+
+    std::string m_description;
+
+    std::set<std::string> m_defaultValue;
```
mathbagu, line 72: Default value should be a `std::string`

include/powsybl/iidm/converter/xml/Constants.hpp
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright (c) 2019, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#ifndef POWSYBL_IIDM_CONVERTER_CONSTANTS_HPP
+#define POWSYBL_IIDM_CONVERTER_CONSTANTS_HPP
```
mathbagu, line 9: Guardian is not OK: It's already use by the Constants.hpp in converter. Maybe you should move this constant somewhere else (ImportOptions and ExporterOptions)?

src/iidm/Identifiable.cpp
```diff
@@ -6,6 +6,7 @@
  */
 
 #include <powsybl/iidm/Identifiable.hpp>
+#include <powsybl/stdcxx/exception.hpp>
```
mathbagu, line 9: Unused: it seems this branch is not up-to-date

src/iidm/Network.cpp
```diff
@@ -43,6 +49,15 @@ namespace powsybl {
...
+static const converter::Parameter ONLY_MAIN_CC_PARAMETER(converter::xml::ONLY_MAIN_CC, converter::Parameter::ParameterType::BOOLEAN, "Export only main CC", false);
+static const converter::Parameter THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER = std::move(converter::Parameter(converter::xml::THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND, converter::Parameter::ParameterType::BOOLEAN, "Throw exception if extension not found", false).addAdditionalNames({"throwExceptionIfExtensionNotFound"}));
+static const converter::Parameter TOPOLOGY_LEVEL_PARAMETER(converter::xml::TOPOLOGY_LEVEL, converter::Parameter::ParameterType::STRING, "Export network in this topology level", std::string("NODE_BREAKER"));
+static const converter::Parameter VERSION_PARAMETER(converter::xml::VERSION, converter::Parameter::ParameterType::STRING, "IIDM-XML version in which files will be generated", converter::xml::IidmXmlVersion::CURRENT_IIDM_XML_VERSION().toString("."));
+static const converter::Parameter WITH_BRANCH_STATE_VARIABLES_PARAMETER(converter::xml::WITH_BRANCH_STATE_VARIABLES, converter::Parameter::ParameterType::BOOLEAN, "Export network with branch state variables", true);
```
mathbagu, line 59: Maybe you should move these options in ImportOptions and ExportOptions.

A single constructor could be OK, if it's you use a sting parameter for the default value.

src/iidm/Network.cpp
```diff
@@ -56,6 +71,13 @@ Network Network::readXml(std::istream& istream, const converter::ImportOptions&
     return converter::xml::NetworkXml::read(istream, options, anonymizer);
 }
 
+Network Network::readXml(std::istream& istream, const stdcxx::Properties& parameters, const converter::Anonymizer& anonymizer) {
+    converter::ImportOptions options;
+    options.setThrowExceptionIfExtensionNotFound(converter::ConversionParameters::readBooleanParameter(parameters, THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER))
+        .setExtensions(converter::ConversionParameters::readStringListParameter(parameters, EXTENSIONS_LIST_PARAMETER));
```
mathbagu, line 77: `converter::ImportOptions options(properties);` whould be nicer.

src/iidm/Network.cpp
```diff
@@ -65,6 +87,19 @@ std::unique_ptr<converter::Anonymizer> Network::writeXml(std::ostream& ostream,
...
+std::unique_ptr<converter::Anonymizer> Network::writeXml(std::ostream& ostream, const Network& network, const stdcxx::Properties& parameters) {
+    converter::ExportOptions options;
+    options.setIndent(converter::ConversionParameters::readBooleanParameter(parameters, INDENT_PARAMETER))
+        .setWithBranchSV(converter::ConversionParameters::readBooleanParameter(parameters, WITH_BRANCH_STATE_VARIABLES_PARAMETER))
+        .setOnlyMainCc(converter::ConversionParameters::readBooleanParameter(parameters, ONLY_MAIN_CC_PARAMETER))
+        .setAnonymized(converter::ConversionParameters::readBooleanParameter(parameters, ANONYMISED_PARAMETER))
+        .setTopologyLevel(Enum::fromString<TopologyLevel>(converter::ConversionParameters::readStringParameter(parameters, TOPOLOGY_LEVEL_PARAMETER)))
+        .setThrowExceptionIfExtensionNotFound(converter::ConversionParameters::readBooleanParameter(parameters, THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND_PARAMETER))
+        .setExtensions(converter::ConversionParameters::readStringListParameter(parameters, EXTENSIONS_LIST_PARAMETER))
+        .setVersion(converter::ConversionParameters::readStringParameter(parameters, VERSION_PARAMETER));
```
mathbagu, line 99: Same remark about the constructor

src/iidm/converter/Parameter.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+Parameter::Parameter(std::string&& name, const ParameterType& type, std::string&& description, std::set<std::string>&& defaultValue) :
+    m_type(type),
+    m_description(std::move(description)),
+    m_defaultValue(std::move(defaultValue)) {
+    m_names.emplace_back(std::move(name));
+}
+
+Parameter& Parameter::addAdditionalNames(const std::initializer_list<std::string>& additionalNames) {
+    std::copy(additionalNames.begin(), additionalNames.end(), std::back_inserter(m_names));
```
mathbagu, line 40: Could you use std::move instead of std::copy?

include/powsybl/iidm/converter/Parameter.hpp
sebalaig: You suggested it here https://github.com/powsybl/powsybl-iidm4cpp/pull/106#discussion_r430050939.
I applied the modification, but forgot to ask you why you wanted to delete it => reverted

include/powsybl/iidm/converter/xml/Constants.hpp
sebalaig: You're right, the guard should be 
```
#ifndef POWSYBL_IIDM_CONVERTER_XML_CONSTANTS_HPP
```

as this file is in include\powsybl\iidm\converter\xml

src/iidm/converter/Parameter.cpp
sebalaig: Yes it's ok with `std::move()`

include/powsybl/iidm/Network.hpp
sebalaig: I agree with creating constructors which take a `stdcxx::Properties` as parameter, but it adds a dependency between `converter` and `converter::xml` module. Can we accept it (related to https://github.com/powsybl/powsybl-iidm4cpp/pull/106#discussion_r435832078)?

PR #107, opened 2020-05-19
Add ActivePowerControlAdder and use it from xml serializer

sebalaig:

Issue #14 

**Other information**:

### DIFF
--- a/extensions/iidm/src/ActivePowerControlAdder.cpp
+++ b/extensions/iidm/src/ActivePowerControlAdder.cpp
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <powsybl/iidm/extensions/iidm/ActivePowerControlAdder.hpp>
+
+#include <powsybl/iidm/Battery.hpp>
+#include <powsybl/iidm/Generator.hpp>
+#include <powsybl/iidm/extensions/iidm/ActivePowerControl.hpp>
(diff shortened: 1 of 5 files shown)


extensions/iidm/src/ActivePowerControlAdder.cpp
```diff
@@ -0,0 +1,53 @@
...
+ActivePowerControlAdder::ActivePowerControlAdder(Extendable& extendable) :
+    ExtensionAdder(extendable) {
+}
+
+std::unique_ptr<Extension> ActivePowerControlAdder::createExtension(Extendable& extendable) {
+    if (stdcxx::isInstanceOf<Battery>(extendable)) {
+        return stdcxx::make_unique<ActivePowerControl>(dynamic_cast<Battery&>(extendable), m_participate, m_droop);
+    }
+    if (stdcxx::isInstanceOf<Generator>(extendable)) {
+        return stdcxx::make_unique<ActivePowerControl>(dynamic_cast<Generator&>(extendable), m_participate, m_droop);
```
mathbagu, line 32: I think it should not be possible anymore to create an ActivePowerControl extension without using the adder. So I propose to make the constructor private, and declare the ActivePowerControlAdder as a friend class.
@sebalaig What do you think?

I think you have to cleanup the existing code, specially in unit tests to use the new API.

sebalaig: Will adders be available for `ActivePowerControl` and `Xnode` extensions only? So, syntax for creating an extension would depend on the extension the developper wants to create ? IMHO this would not be very convenient/intuitive.

mathbagu: All the extension will be migrate to follow the new design. But we do not have enough time to do it in Java, so few of them support the new design.

sebalaig: We decided to be Java consistent and to enable the creation without using the adder.

Issue #36, opened 2020-03-03
[IIDM v1.1] Migration of XNode extension to the new design


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature
https://github.com/powsybl/powsybl-core/pull/1153
https://github.com/powsybl/powsybl-core/pull/1091

* **What is the current behavior?**
The XNode extension use the old design.

* **What is the expected behavior?**
The XNode extension should use the new design of extensions.

* **What is the motivation / use case for changing the behavior?**
Have both Java and C++ API as close as possible

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #108, opened 2020-05-19
Add XnodeAdder and use it from xml serializer

sebalaig:

Issue #36 

**Other information**:

### DIFF
--- a/extensions/entsoe/src/XnodeAdder.cpp
+++ b/extensions/entsoe/src/XnodeAdder.cpp
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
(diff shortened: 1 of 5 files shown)


extensions/entsoe/src/XnodeAdder.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+namespace entsoe {
+
+XnodeAdder::XnodeAdder(Extendable& extendable) :
+    ExtensionAdder(extendable) {
+}
+
+std::unique_ptr<Extension> XnodeAdder::createExtension(Extendable& extendable) {
+    if (stdcxx::isInstanceOf<DanglingLine>(extendable)) {
+        return stdcxx::make_unique<Xnode>(dynamic_cast<DanglingLine&>(extendable), m_code);
```
mathbagu, line 29: I think it should not be possible anymore to create an Xnode extension without using the adder. So I propose to make the constructor private, and declare the XNodeAdder as a friend class.
@sebalaig What do you think?

I think you have to cleanup the existing code, specially in unit tests to use the new API.

sebalaig: We decided to be Java consistent and to enable the creation without using the adder.

Issue #62, opened 2020-03-23
[IIDM v1.2] Any Identifiable can be fictitious, not only switches


MioRtia:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
Switches can be fictitious.

* **What is the expected behavior?**
Any identifiable can be fictitious.

* **What is the motivation / use case for changing the behavior?**
Any equipment can be the equivalent of a part of a network.

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

Linked to PR powsybl/powsybl-core#1209

PR #117, opened 2020-06-10
[IIDM v1.2] Any Identifiable can be fictitious, not only switches (#62)

sebalaig:

Issue #62 

Feature

Only switches can be fictitious.

Every identifiable can be fictitious.

**Other information**:


sebalaig:
@mathbagu 
There 1 thing to handle in this PR : from now, as `Identifiable` is not templated, user cannot write : 
```
    Substation& s = ...
    s.setFictitious(true).setCountry(Country::FR);
```
as setFictitious returns an `Identifiable&`.

2 possibilities to handle that : 
 - Make Identifiable/Connectable/Branch templated : I could remerber we didn't want to do that to make API fluent; also there was a problem with terminals stored somewhere but maybe I'm wrong :/
 - Overriding setFictitious on every base class to use covariant return type (which is done for switches and voltage levels)

I propose to implement the second one. If you have any other solution, I'm interested ;)

sebalaig:
> @mathbagu
> There 1 thing to handle in this PR : from now, as `Identifiable` is not templated, user cannot write :
> 
> ```
>     Substation& s = ...
>     s.setFictitious(true).setCountry(Country::FR);
> ```
> 
> as setFictitious returns an `Identifiable&`.
> 
> 2 possibilities to handle that :
> 
> * Make Identifiable/Connectable/Branch templated : I could remerber we didn't want to do that to make API fluent; also there was a problem with terminals stored somewhere but maybe I'm wrong :/
> * Overriding setFictitious on every base class to use covariant return type (which is done for switches and voltage levels)
> 
> I propose to implement the second one. If you have any other solution, I'm interested ;)

We decided to use covariant return type as there is no other simple solution.

Issue #122, opened 2020-06-18
[IIDM v1.2] Remove skip-extension property


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature
https://github.com/powsybl/powsybl-core/pull/1212

We should ensure that it's possible to read an XIIDM file that contains extension and to write it as an XML file without extensions

sebalaig:
Fixed in PR #136

PR #136, opened 2020-06-24
[IIDM v1.2] Fix serialization with no exported extensions

sebalaig:
- [x] Docs have been added / updated (for bug fixes / features)

Related to #122 

Bug fix

When user configure to export no extension in XML output, <extension> tag is generated for this identifiable, but have no child.

When user configure to export no extension in XML output, <extension> tag is not generated for this identifiable anymore (like in Java).

**Other information**:


mathbagu:
Could you please rebase this branch to make the review easier?

include/powsybl/iidm/converter/ExportOptions.hpp
```diff
@@ -110,6 +110,15 @@ class ExportOptions {
...
     const TopologyLevel& getTopologyLevel() const;
 
+    /**
+     * Return true if one (or more) extension of the given list should be exported
+     *
+     * @param extensions the list of extensions to be checked.
+     *
+     * @return true if at least one extension should be exported, false otherwise
+     */
+    bool hasAtLeastOneExtension(const std::set<std::string>& extensions) const;
```
mathbagu, line 120: Looking to the java code, there is an `AbstractOptions` class to share code between `ImportOptions` and `ExportOptions`. Even it's not used, I propose to refactor the option classes to make the code more similar to the java.

src/iidm/converter/xml/NetworkXml.cpp
```diff
@@ -174,20 +174,28 @@ void writeExtension(const Extension& extension, NetworkXmlWriterContext& context
...
 }
 
+std::set<std::string> getExtensionsName(const stdcxx::const_range<Extension>& extensions) {
+    std::set<std::string> names;
+    for (const auto& extension : extensions) {
+        names.emplace(extension.getName());
+    }
+    return names;
+}
+
```
mathbagu, line 184: I think you should use a range transformation instead of creating a set.

test/iidm/converter/xml/SkipExtensionTest.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace iidm {
+
+namespace converter {
+
+namespace xml {
+
+BOOST_AUTO_TEST_SUITE(SkipExtensionTestSuite)
+
+BOOST_FIXTURE_TEST_CASE(SkipExtensionTest, test::ResourceFixture) {
```
mathbagu, line 29: Is there a unit test to check if skiping extension during the read is working well?

test/iidm/converter/xml/SkipExtensionTest.cpp
```diff
@@ -0,0 +1,50 @@
...
+    Network network = Network::readXml(test::converter::RoundTrip::getVersionedNetwork("multiple-extensions.xml", IidmXmlVersion::V1_0()));
+    const std::string& refNetwork = test::converter::RoundTrip::getVersionedNetwork("noExtension.xml", IidmXmlVersion::V1_0());
+
+    stdcxx::Properties properties;
+    std::stringstream ostream;
+
+    properties.set(ExportOptions::EXTENSIONS_LIST, "");
```
mathbagu, line 36: Does it work properly if this property is omitted?

include/powsybl/iidm/converter/ExportOptions.hpp
sebalaig: Postponed. I created the issue #140 to report the fix.

Issue #64, opened 2020-03-23
[IIDM v1.2] Change mechanism for creating and cleaning vertices in a node/breaker voltage level's graph and delete nodeCount attribute in nodeBreakerTopology XML element


MioRtia:
* **Do you want to request a *feature* or report a *bug*?**
Both

* **What is the current behavior?**
The node count can be set in node breaker voltage level. All nodes are pre-created. When an equipment is removed, the vertex is associated to null.

* **What is the expected behavior?**
The node count should be set. It should be updated each time an equipment is attached in the voltage level. All nodes with no equipment attached should not exist or be null.

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

Linked to PR powsybl/powsybl-core#1161

sebalaig:
Fixed in PR #137

PR #137, opened 2020-07-01
[IIDM v1.2] Update node creation mechanism

sebalaig:

Issue #64 

- [x] The *Breaking Change* or *Deprecated* label has been added

**Other information**:

### DIFF
--- a/include/powsybl/iidm/VoltageLevelViews.hpp
+++ b/include/powsybl/iidm/VoltageLevelViews.hpp
@@ -127,8 +127,6 @@ public:
 
     virtual unsigned long getNode2(const std::string& switchId) const = 0;
 
-    virtual unsigned long getNodeCount() const = 0;
-
     virtual stdcxx::const_range<unsigned long> getNodes() const = 0;
 
     virtual stdcxx::CReference<Switch> getSwitch(const std::string& switchId) const = 0;
--- a/include/powsybl/iidm/VoltageLevelViews.hpp
+++ b/include/powsybl/iidm/VoltageLevelViews.hpp
@@ -121,6 +121,8 @@ public:
 
     virtual stdcxx::const_range<InternalConnection> getInternalConnections() const = 0;
 
+    virtual unsigned long getMaximumNodeIndex() const = 0;
+
     virtual unsigned long getNode1(const std::string& switchId) const = 0;
 
     virtual unsigned long getNode2(const std::string& switchId) const = 0;
@@ -161,9 +163,9 @@ public:
 
     virtual SwitchAdder newSwitch() = 0;
 
-    virtual void removeSwitch(const std::string& switchId) = 0;
+    virtual void removeInternalConnections(unsigned long node1, unsigned long node2) = 0;
 
-    virtual NodeBreakerView& setNodeCount(unsigned long nodeCount) = 0;
+    virtual void removeSwitch(const std::string& switchId) = 0;
 
     virtual void traverse(unsigned long node, const Traverser& traverser) const = 0;
 };
--- a/include/powsybl/math/UndirectedGraph.hxx
+++ b/include/powsybl/math/UndirectedGraph.hxx
@@ -40,7 +40,7 @@ void UndirectedGraph<V, E>::checkVertex(unsigned long v) const {
 
 template <typename V, typename E>
 void UndirectedGraph<V, E>::cleanVertices(unsigned long v) {
-    for (unsigned long i = v; v >= 0; --i) {
+    for (unsigned long i = v; i >= 0; --i) {
         const auto& it = m_availableVertices.find(i);
         if (it == m_availableVertices.end()) {
             return;
--- a/include/powsybl/math/UndirectedGraph.hxx
+++ b/include/powsybl/math/UndirectedGraph.hxx
@@ -39,6 +39,18 @@ void UndirectedGraph<V, E>::checkVertex(unsigned long v) const {
 }
 
 template <typename V, typename E>
+void UndirectedGraph<V, E>::cleanVertices(unsigned long v) {
+    for (unsigned long i = v; v >= 0; --i) {
+        const auto& it = m_availableVertices.find(i);
+        if (it == m_availableVertices.end()) {
+            return;
+        }
+        m_availableVertices.erase(it);
+        m_vertices.pop_back();
+    }
+}
+
+template <typename V, typename E>
 unsigned long UndirectedGraph<V, E>::addEdge(unsigned long v1, unsigned long v2, const stdcxx::Reference<E>& object) {
     checkVertex(v1);
     checkVertex(v2);
@@ -66,12 +78,12 @@ unsigned long UndirectedGraph<V, E>::addVertex() {
     unsigned long v;
 
     std::unique_ptr<Vertex> vertex = stdcxx::make_unique<Vertex>();
-    if (m_removedVertices.empty()) {
+    if (m_availableVertices.empty()) {
         v = m_vertices.size();
         m_vertices.emplace_back(std::move(vertex));
     } else {
-        v = *m_removedVertices.begin();
-        m_removedVertices.erase(m_removedVertices.begin());
+        v = *m_availableVertices.begin();
+        m_availableVertices.erase(m_availableVertices.begin());
 
         m_vertices[v] = std::move(vertex);
     }
@@ -79,6 +91,25 @@ unsigned long UndirectedGraph<V, E>::addVertex() {
     invalidateAdjacencyList();
 
     return v;
+}
+
+template <typename V, typename E>
+void UndirectedGraph<V, E>::addVertexIfNotPresent(unsigned long v) {
+    if (v < m_vertices.size()) {
+        const auto& it = m_availableVertices.find(v);
+        if (it != m_availableVertices.end()) {
+            m_vertices[v] = stdcxx::make_unique<Vertex>();
+            m_availableVertices.erase(it);
+        }
+    } else {
+        std::size_t oldSize = m_vertices.size();
+        m_vertices.resize(v + 1);
+        for (unsigned int i = oldSize; i < v; i++) {
+            m_availableVertices.insert(m_availableVertices.end(), i);
+        }
+        m_vertices[v] = stdcxx::make_unique<Vertex>();
+    }
+    invalidateAdjacencyList();
 }
 
 template <typename V, typename E>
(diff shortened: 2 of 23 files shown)


include/powsybl/iidm/VoltageLevelViews.hpp
```diff
@@ -161,9 +163,9 @@ class NodeBreakerView {
 
     virtual SwitchAdder newSwitch() = 0;
 
-    virtual void removeSwitch(const std::string& switchId) = 0;
+    virtual void removeInternalConnections(unsigned long node1, unsigned long node2) = 0;
 
-    virtual NodeBreakerView& setNodeCount(unsigned long nodeCount) = 0;
+    virtual void removeSwitch(const std::string& switchId) = 0;
```
mathbagu, line 168: You forgot to remove getNodeCount(): if you want to manage backward compatibility, add a default behavior and make this method not virtual

include/powsybl/math/UndirectedGraph.hxx
```diff
@@ -81,6 +93,25 @@ unsigned long UndirectedGraph<V, E>::addVertex() {
...
+void UndirectedGraph<V, E>::addVertexIfNotPresent(unsigned long v) {
+    if (v < m_vertices.size()) {
+        const auto& it = m_availableVertices.find(v);
+        if (it != m_availableVertices.end()) {
+            m_vertices[v] = stdcxx::make_unique<Vertex>();
+            m_availableVertices.erase(it);
+        }
+    } else {
+        std::size_t oldSize = m_vertices.size();
+        m_vertices.resize(v + 1);
```
mathbagu, line 106: Moving this line after the for loop will allow you to remove the oldSize variable, and write:
```
        for (unsigned int i = m_vertices.size(); i < v; ++i) {
            m_availableVertices.insert(m_availableVertices.end(), i);
        }
        m_vertices.resize(v + 1);
```
I don't understand why you try to add the value at the end of the set, using a position, and not simply insert values in the set: `m_availableVertices.insert(i);`

include/powsybl/math/UndirectedGraph.hxx
```diff
@@ -38,6 +38,18 @@ void UndirectedGraph<V, E>::checkVertex(unsigned long v) const {
     }
 }
 
+template <typename V, typename E>
+void UndirectedGraph<V, E>::cleanVertices(unsigned long v) {
```
mathbagu, line 42: I think this method is not well placed. The `checkVertex` and `checkEdge` are at the top of the file cause they are used everywhere, and you need the implementation before calling them. The cleanVertices is used only in removeVertex, so it's OK to move this method.

src/iidm/NodeBreakerVoltageLevel.cpp
```diff
@@ -325,6 +338,22 @@ void NodeBreakerVoltageLevel::reduceVariantArraySize(unsigned long number) {
     m_variants.reduceVariantArraySize(number);
 }
 
+void NodeBreakerVoltageLevel::removeInternalConnections(unsigned long node1, unsigned long node2) {
+    const auto& filter = [this, node1, node2](const unsigned long& edge) {
+        return !m_graph.getEdgeObject(edge) && m_graph.getVertex1(edge) == node1 && m_graph.getVertex2(edge) == node2;
```
mathbagu, line 343: This is not equivalent to the java code. An internal connection is not an oriented edge, so you also have to check if
`m_graph.getVertex1(edge) == node2 && m_graph.getVertex2(edge) == node1`

src/iidm/NodeBreakerVoltageLevel.cpp
```diff
@@ -325,6 +338,22 @@ void NodeBreakerVoltageLevel::reduceVariantArraySize(unsigned long number) {
...
+    };
+
+    const auto& internalConnectionsToBeRemoved = m_graph.getEdges() | boost::adaptors::filtered(filter);
+    if (boost::size(internalConnectionsToBeRemoved) == 0UL) {
```
mathbagu, line 347: Use `boost::empty` instead

src/iidm/NodeBreakerVoltageLevel.cpp
```diff
@@ -336,13 +365,6 @@ void NodeBreakerVoltageLevel::removeSwitch(const std::string& switchId) {
     getNetwork().remove(aSwitch.get());
 }
 
-void NodeBreakerVoltageLevel::setNodeCount(unsigned long nodeCount) {
-    unsigned long oldCount = m_graph.getVertexCount();
-    for (unsigned long i = oldCount; i < nodeCount; ++i) {
-        m_graph.addVertex();
-    }
-}
```
mathbagu, line 344: Some changes are missing: the `attach` method has not been updated

src/iidm/NodeBreakerVoltageLevel.hpp
```diff
@@ -105,8 +114,18 @@ class NodeBreakerVoltageLevel : public VoltageLevel {
 
     unsigned long getNode2(const std::string& switchId) const;
 
+    /**
+     * @deprecated Use {@link #getMaximumNodeIndex()} instead.
+     */
     unsigned long getNodeCount() const;
```
mathbagu, line 120: If the method is private and deprecated, why do you keep it?

src/iidm/NodeBreakerVoltageLevel.hpp
```diff
@@ -70,6 +75,8 @@ class NodeBreakerVoltageLevel : public VoltageLevel {
 
     bool isConnected(const Terminal& terminal) const;
 
+    void removeInternalConnections(unsigned long node1, unsigned long node2);
```
mathbagu, line 78: This method is not well placed: it should not be in the public API. In java, this method is in the different views. If this method is used by the view, as the view is a friend class, move this class in the private part

src/iidm/NodeBreakerVoltageLevel.hpp
```diff
@@ -105,8 +114,18 @@ class NodeBreakerVoltageLevel : public VoltageLevel {
...
+     * @deprecated Use {@link #getMaximumNodeIndex()} instead.
+     */
     unsigned long getNodeCount() const;
 
+    /**
+     * Get the list of nodes.
+     *
+     * @return the list of nodes.
+     */
+    stdcxx::const_range<unsigned long> getNodes() const;
```
mathbagu, line 127: Same remark there

src/iidm/NodeBreakerVoltageLevelViews.cpp
```diff
@@ -206,7 +210,7 @@ unsigned long NodeBreakerViewImpl::getNodeCount() const {
 }
 
 stdcxx::const_range<unsigned long> NodeBreakerViewImpl::getNodes() const {
-    return m_voltageLevel.getGraph().getVertices();
+    return m_voltageLevel.getNodes();
```
mathbagu, line 213: I don't understand this change: the implementation will differ from the java code

src/iidm/NodeBreakerVoltageLevelViews.cpp
```diff
@@ -285,13 +289,12 @@ NodeBreakerViewImpl::SwitchAdder NodeBreakerViewImpl::newSwitch() {
     return SwitchAdder(m_voltageLevel);
 }
 
-void NodeBreakerViewImpl::removeSwitch(const std::string& switchId) {
-    m_voltageLevel.removeSwitch(switchId);
+void NodeBreakerViewImpl::removeInternalConnections(unsigned long node1, unsigned long node2) {
+    m_voltageLevel.removeInternalConnections(node1, node2);
```
mathbagu, line 293: Maybe the removeInternalConnections implementation should be moved here: you only need the graph?

src/iidm/converter/xml/VoltageLevelXml.cpp
```diff
@@ -91,8 +93,10 @@ void VoltageLevelXml::readCalculatedBus(VoltageLevel &voltageLevel, NetworkXmlRe
 }
 
 void VoltageLevelXml::readNodeBreakerTopology(VoltageLevel& voltageLevel, NetworkXmlReaderContext& context) const {
-    const auto& nodeCount = context.getReader().getAttributeValue<unsigned long>(NODE_COUNT);
-    voltageLevel.getNodeBreakerView().setNodeCount(nodeCount);
+    IidmXmlUtil::runUntilMaximumVersion(IidmXmlVersion::V1_1(), context.getVersion(), []() {
+        logging::Logger& logger = logging::LoggerFactory::getLogger<VoltageLevelXml>();
+        logger.info(stdcxx::format("attribute %1%.nodeCount is ignored.", NODE_BREAKER_TOPOLOGY));
+    });
```
mathbagu, line 99: This log has been removed in https://github.com/powsybl/powsybl-core/pull/1299 because too noisy

include/powsybl/math/UndirectedGraph.hxx
sebalaig: Agree... I though it was more optimized, but I forgot that we are pushing decreasingly :sweat_smile:

src/iidm/NodeBreakerVoltageLevel.cpp
```diff
@@ -52,6 +55,7 @@ void NodeBreakerVoltageLevel::attach(Terminal& terminal, bool test) {
     if (!test) {
         auto& nodeTerminal = dynamic_cast<NodeTerminal&>(terminal);
         unsigned long node = nodeTerminal.getNode();
+        m_graph.addVertexIfNotPresent(node);
```
sebalaig, line 58: https://github.com/powsybl/powsybl-iidm4cpp/pull/137#discussion_r451380468
Yes is is here.

src/iidm/NodeBreakerVoltageLevel.cpp
sebalaig: Yes it does : https://github.com/powsybl/powsybl-iidm4cpp/pull/137/files#r451423417

src/iidm/NodeBreakerVoltageLevelViews.cpp
sebalaig: Ok reverted.

src/iidm/NodeBreakerVoltageLevel.hpp
sebalaig: Ok removed

src/iidm/NodeBreakerVoltageLevelViews.cpp
sebalaig: removeInternalConnections modifies the graph. Is the view allowed to do that ?

include/powsybl/iidm/VoltageLevelViews.hpp
sebalaig: Ok removed. To be documented in V1.2.0 release note

include/powsybl/math/UndirectedGraph.hxx
sebalaig: >The checkVertex and checkEdge are at the top of the file cause they are used everywhere

Strange! Without using a hxx file, I would understand but here I don't. I moved them and its ok.

src/iidm/converter/xml/VoltageLevelXml.cpp
sebalaig: Ok removed also

Issue #75, opened 2020-04-02
[IIDM v1.2] Add convenient methods in VoltageLevel.NodeBreakerView


MioRtia:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
There is only one method `getTerminal(int node)` in `VoltageLevel.NodeBreakerView` which returns a `Terminal` associated with `node` if it exists, `null` if this node is only associated with a `Switch` or an `InternalConnection` and throws an exception if no equipment is attached to this node.

* **What is the expected behavior?**
We want a `getOptionalTerminal(int node)` method which returns an `Optional<Terminal>` with the linked `Terminal` if it exists or an empty `Optional` if the node is associated with a `Switch`, an `InternalConnection` or has no attached equipment. Furthermore, we want a `hasAttachedEquipment(int node)` which checks if `node` has equipment attached to it.

* **What is the motivation / use case for changing the behavior?**
After the change of building of node/breaker topology, no "empty" vertex is created anymore: if no equipment is attached to a node, the vertex is null, not empty. We want to prevent throwing exceptions in this case.

* **Please tell us about your environment:**
  - PowSyBl Version: 3.3.0-SNAPSHOT

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)
Linked to powsybl/powsybl-core#1235 and powsybl/powsybl-core#1233

sebalaig:
Fixed in PR #141

Issue #67, opened 2020-03-24
Add a convenient function to get a Terminal from its node


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature
https://github.com/powsybl/powsybl-core/pull/1235

* **What is the current behavior?**
The `getTerminal(int node)` method of the NodeBreakerView throw an exception if the node doesn't exist.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
Another method should be added to get an optional of Terminal if the node doesn't exists.

* **What is the motivation / use case for changing the behavior?**
Have both Java and C++ implementation as close as possible.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

sebalaig:
Fixed in PR #141

PR #141, opened 2020-07-09
[IIDM V1.2] Add a convenient function to get a Terminal from its node (#67+#75)

sebalaig:

#67 and #75 

**Other information**:

### DIFF
--- a/include/powsybl/iidm/VoltageLevelViews.hpp
+++ b/include/powsybl/iidm/VoltageLevelViews.hpp
@@ -153,6 +153,8 @@ public:
 
     virtual stdcxx::Reference<Terminal> getTerminal2(const std::string& switchId) = 0;
 
+    virtual bool hasAttachedEquipment(unsigned long node) const = 0;
+
     virtual SwitchAdder newBreaker() = 0;
 
     virtual BusbarSectionAdder newBusbarSection() = 0;
--- a/include/powsybl/iidm/VoltageLevelViews.hpp
+++ b/include/powsybl/iidm/VoltageLevelViews.hpp
@@ -129,6 +129,8 @@ public:
 
     virtual stdcxx::const_range<unsigned long> getNodes() const = 0;
 
+    virtual stdcxx::Reference<Terminal> getOptionalTerminal(unsigned long node) const = 0;
+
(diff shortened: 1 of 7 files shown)


include/powsybl/iidm/VoltageLevelViews.hpp
```diff
@@ -129,6 +129,8 @@ class NodeBreakerView {
 
     virtual stdcxx::const_range<unsigned long> getNodes() const = 0;
 
+    virtual stdcxx::Reference<Terminal> getOptionalTerminal(unsigned long node) const = 0;
```
mathbagu, line 132: You have to create two variants of this method, one `const` that returns a `CReference` and one no-const that returns a `Reference`

include/powsybl/math/UndirectedGraph.hxx
```diff
@@ -450,6 +450,14 @@ void UndirectedGraph<V, E>::traverse(unsigned long v, const Traverser& traverser
     }
 }
 
+template <typename V, typename E>
+bool UndirectedGraph<V, E>::vertexExists(unsigned long v) const {
+    if (v >= m_vertices.size()) {
+        throw PowsyblException(stdcxx::format("Invalid vertex %1%", v));
+    }
+    return static_cast<bool>(m_vertices[v]);
+}
```
mathbagu, line 459: This is not consistent with the java code: In java, if `v` is greater or equal to `m_vertices.size()`, it returns false, it doesn't throw an exception.
As C++ makes lazy evaluation, you can simply copy/paste the java code

sebalaig: `v` is an `unsigned long` so the method cannot throw any exception unlike in Java.

include/powsybl/iidm/VoltageLevelViews.hpp
sebalaig: Ok, done

include/powsybl/math/UndirectedGraph.hxx
mathbagu: Agree, but in your first version you would throw an exception if `v` is greater than the vertices size. This was my point. The java code, in that specific case returned false. I see that in the last version you fix it.

Issue #125, opened 2020-06-18
[IIDM v1.2] Fix NullPointerException when removing a voltage level with internal connections


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Bug
https://github.com/powsybl/powsybl-core/pull/1259

In this PR, we fix an issue concerning the removal of a voltage level with internal connections

sebalaig:
Fixed in PR #145

PR #145, opened 2020-07-15
Fix crash when removing a VoltageLevel

sebalaig:

No (found while testing issue #125)

Bug fix

Removing a VoltageLevel from a network crashes because `NetworkIndex::m_objectsByType` contains 1 entry/key for every VoltageLevels in the Network. While removing a NodeBreakerVoltageLevel, the effective type is not found in the map => crash

Now, `NetworkIndex::m_objectsByType` contains 1 entry/key for each concrete VoltageLevel (Node Breaker / Bus Breaker) in the network.
Accessors `NetworkIndex::getAll` and `NetworkIndex::getObjectCount` are now specialized for `VoltageLevel` to concatenate concrete VLs.

**Other information**:

### DIFF
--- a/include/powsybl/iidm/NetworkIndex.hxx
+++ b/include/powsybl/iidm/NetworkIndex.hxx
@@ -41,7 +41,7 @@ T& NetworkIndex::checkAndAdd(std::unique_ptr<T>&& identifiable) {
     auto it = m_objectsById.emplace(std::make_pair(identifiable->getId(), std::move(identifiable)));
 
     std::reference_wrapper<Identifiable> refIdentifiable = *it.first->second;
-    m_objectsByType[typeid(T)].emplace_back(refIdentifiable);
+    m_objectsByType[typeid(*it.first->second)].emplace_back(refIdentifiable);
 
     return dynamic_cast<T&>(*it.first->second);
 }
@@ -79,6 +79,12 @@ stdcxx::const_range<Identifiable> NetworkIndex::getAll<Identifiable, Identifiabl
 
 template <>
 stdcxx::range<Identifiable> NetworkIndex::getAll<Identifiable, Identifiable>();
+
+template <>
+stdcxx::const_range<VoltageLevel> NetworkIndex::getAll<VoltageLevel, VoltageLevel>() const;
+
+template <>
+stdcxx::range<VoltageLevel> NetworkIndex::getAll<VoltageLevel, VoltageLevel>();
 
 template <typename T, typename U>
 stdcxx::const_range<U> NetworkIndex::getAll() const {
--- a/include/powsybl/iidm/NetworkIndex.hxx
+++ b/include/powsybl/iidm/NetworkIndex.hxx
@@ -41,7 +41,7 @@ T& NetworkIndex::checkAndAdd(std::unique_ptr<T>&& identifiable) {
     auto it = m_objectsById.emplace(std::make_pair(identifiable->getId(), std::move(identifiable)));
 
     std::reference_wrapper<Identifiable> refIdentifiable = *it.first->second;
-    m_objectsByType[typeid(*it.first->second)].emplace_back(refIdentifiable);
+    m_objectsByType[typeid(T)].emplace_back(refIdentifiable);
 
     return dynamic_cast<T&>(*it.first->second);
 }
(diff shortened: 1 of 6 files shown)


include/powsybl/iidm/NetworkIndex.hxx
```diff
@@ -41,7 +41,7 @@ T& NetworkIndex::checkAndAdd(std::unique_ptr<T>&& identifiable) {
     auto it = m_objectsById.emplace(std::make_pair(identifiable->getId(), std::move(identifiable)));
 
     std::reference_wrapper<Identifiable> refIdentifiable = *it.first->second;
-    m_objectsByType[typeid(T)].emplace_back(refIdentifiable);
+    m_objectsByType[typeid(*it.first->second)].emplace_back(refIdentifiable);
```
mathbagu, line 44: I would change the caller in `VoltageLevelAdder` instead.

test/iidm/NodeBreakerVoltageLevelTest.cpp
```diff
@@ -177,6 +180,109 @@ Network createCalculatedBusSwitchTestNetwork() {
...
 }
 
+void addInternalConnection( VoltageLevel::NodeBreakerView& topo, int node1, int node2) {
+    topo.newInternalConnection()
+        .setNode1(node1)
+        .setNode2(node2)
+        .add();
+}
+
+const std::string S5_10K_V = "S5 10kV";
```
mathbagu, line 190: That's uggly... create a local variable in createNetwork is better!

test/iidm/NodeBreakerVoltageLevelTest.cpp
```diff
@@ -1046,6 +1152,15 @@ BOOST_AUTO_TEST_CASE(TerminalTest) {
...
 }
 
+BOOST_AUTO_TEST_CASE(TestRemoveVoltageLevelWithInternalConnectionsIssue) {
+    Network network("test", "test");
+
+    createNetwork(network);
+    network.getLine("L6").remove(); // needed to be allowed to remove the voltage level
+    network.remove(network.getVoltageLevel(S5_10K_V));
+    POWSYBL_ASSERT_THROW(network.getVoltageLevel(S5_10K_V), PowsyblException, "Unable to find to the identifiable 'S5 10kV'");
+}
```
mathbagu, line 1162: I think this test is also relevant for Bus/Breaker Voltage levels

src/iidm/NetworkIndex.cpp
```diff
@@ -81,11 +85,28 @@ stdcxx::range<MultiVariantObject> NetworkIndex::getAll<Identifiable, MultiVarian
...
+stdcxx::const_range<VoltageLevel> NetworkIndex::getAll<VoltageLevel, VoltageLevel>() const {
+    stdcxx::const_range<VoltageLevel> bbvl = getAll<BusBreakerVoltageLevel, VoltageLevel>();
+    return boost::range::join(bbvl, getAll<NodeBreakerVoltageLevel, VoltageLevel>());
+}
+
+template <>
+stdcxx::range<VoltageLevel> NetworkIndex::getAll<VoltageLevel, VoltageLevel>() {
+    stdcxx::range<VoltageLevel> bbvl = getAll<BusBreakerVoltageLevel, VoltageLevel>();
+    return boost::range::join(bbvl, getAll<NodeBreakerVoltageLevel, VoltageLevel>());
+}
```
mathbagu, line 98: Two options: create two local variables or none, but do not implement these methods like this.

test/iidm/NodeBreakerVoltageLevelTest.cpp
sebalaig: Agree, but this variable is used in test `TestRemoveVoltageLevelWithInternalConnectionsIssue`. 
I updated the test to hardcode the VL name, like it is done in every other test :
```
    network.remove(network.getVoltageLevel("S5 10kV"));
    POWSYBL_ASSERT_THROW(network.getVoltageLevel("S5 10kV"), PowsyblException, "Unable to find to the identifiable 'S5 10kV'");
```

test/iidm/NodeBreakerVoltageLevelTest.cpp
sebalaig: Ok added also In Bus/Breaker Voltage level tests.

Issue #130, opened 2020-06-18
[IIDM v1.2] New extension "VoltagePerReactivePowerControl" for StaticVarCompensator


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature
https://github.com/powsybl/powsybl-core/pull/1321

Add new `VoltagePerReactivePowerControl` extension for SVC.

sebalaig:
Fixed by PR #168

PR #168, opened 2020-07-24
[IIDM v1.2] Add voltagePerReactivePowerControl extension (#130)

sebalaig:
Backport of https://github.com/powsybl/powsybl-core/pull/1321

#130 

**Other information**:


extensions/iidm/include/powsybl/iidm/extensions/iidm/VoltagePerReactivePowerControl.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Copy assignment operator
+     */
+    VoltagePerReactivePowerControl& operator=(const VoltagePerReactivePowerControl&) = delete;
+
+    /**
+     * Move assignment operator
+     */
+    VoltagePerReactivePowerControl& operator=(VoltagePerReactivePowerControl&&) = delete;
```
mathbagu, line 58: For other extensions, we don't have all the constructors and operators. It would be weird this extension doesn't have the same behavior than the others. Could you confirm the default/delete constructors/assignment operators are exposed in the same way in the other extensions.
If not, I would prefer you remove these copy constructors/assignment operators, and fix all the extensions in the same time.

extensions/iidm/include/powsybl/iidm/extensions/iidm/VoltagePerReactivePowerControlAdder.hpp
```diff
@@ -0,0 +1,84 @@
...
+     * Creates the VoltagePerReactivePowerControl extension.
+     *
+     * @param extendable the extendable
+     *
+     * @return the extension
+     */
+    std::unique_ptr<Extension> createExtension(Extendable& extendable) override;
+
+private:
+    double m_slope = 0.0;
```
mathbagu, line 73: Why the default value is 0. Should it be nan?

extensions/iidm/include/powsybl/iidm/extensions/iidm/VoltagePerReactivePowerControl.hpp
sebalaig: You're right, they are implcitly declared in other extensions :/
I'll remove declarations in this PR and add them all back in another PR. 

Are every extension movable/copiable/assignable ? 
Or it should be non assignable, like adders : 
```
    XXX(const XXX&) = default;
    XXX(XXX&&) = default;
    XXX& operator=(const XXX&) = delete;
    XXX& operator=(XXX&&) = delete;
```

extensions/iidm/include/powsybl/iidm/extensions/iidm/VoltagePerReactivePowerControlAdder.hpp
sebalaig: I agree, but it's Java behaviour. Do I create an issue ?

sebalaig: Fixed by initializing to `stdcxx::nan()`

mathbagu: Could you show me the code that initialize the slope to 0 in the java code? From my understanding, the value is not initialized in the adder, so it is set to NaN.

sebalaig: It sounds that 0.0d is the default value for double : https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html
But maybe I misunderstood.

mathbagu: Hum... you right... my bad. I prefer revert this fix, and be consistent with the java code even I think it's a mistake.

Issue #173, opened 2020-08-13
[IIDM v1.2] Improve clang-tidy performance during CI


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Improvement

* **What is the current behavior?**
During CI, clang-tidy is run on all files, using the `-DCMAKE_CXX_CLANG_TIDY` option. This operation slows down the CI (35 minutes)

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
If `clang-tidy` analysis is performed only on modified files, the CI will be faster and it makes it possible to merge more PR. To do that, we can use the `clang-tidy 'git diff -U0 HEAD^ --name-only' -p build_dir`  command. This tool required the generation of the compilation database that can be generated using the `CMAKE_EXPORT_COMPILE_COMMANDS` option.

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #178, opened 2020-08-14
Optimize CI by invoking clang tidy only on modified files

sebalaig:

#173 

**Other information**:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,11 +11,6 @@ jobs:
     name: Linux
     runs-on: ubuntu-latest
     steps:
-      - name: Install clang-tidy
-        run: |
-          sudo apt-get install -y clang-tidy-9
-          sudo update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-tidy-9 100
-
       - name: Checkout sources
         uses: actions/checkout@v1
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -24,12 +24,19 @@ jobs:
...
-          -DCMAKE_CXX_CLANG_TIDY=clang-tidy
           -DCMAKE_PREFIX_PATH=$BOOST_ROOT_1_72_0
+          -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
 
       - name: Build
         run: cmake --build $GITHUB_WORKSPACE/build-linux --parallel 3
 
+      - name: Clang Tidy
+        run: |
+          export MODIFIED_FILES=$(git diff -U0 HEAD^ --name-only |grep -E "\.(cpp|hpp|hxx)$")
```
mathbagu, line 35: Do not export the variable and add a space between `|` and `grep`

.github/workflows/ci.yml
```diff
@@ -24,12 +24,19 @@ jobs:
...
+          if [ -n "$MODIFIED_FILES" ]; then
+              clang-tidy $MODIFIED_FILES -p $GITHUB_WORKSPACE/build-linux
+          fi
+
```
mathbagu, line 39: I think clang-tidy check should be moved to the `QA` job. Don't forget to remove the installation of clang-tidy from the linux job too.

mathbagu (changes requested):

.github/workflows/ci.yml
sebalaig: Without exporting variable, CI failed : https://github.com/powsybl/powsybl-iidm4cpp/actions/runs/244562520 :/
Maybe I missed something ?

mathbagu: I don't understand why there is an error. On my Fedora, without the `export`, it works properly. We keep it like that

Issue #186, opened 2020-09-23
[IIDM v1.3] Support double in conversion parameters


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
Conversion parameters only support string, list of string and boolean parameters

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
Support double parameters as implemented in https://github.com/powsybl/powsybl-core/pull/1380

* **What is the motivation / use case for changing the behavior?**
Have both Java an C++ implementation as close as possible.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #195, opened 2020-09-30
[IIDM v1.3] Add support for double parameters (#186)

sebalaig:

#186

**Other information**:


src/iidm/converter/Parameter.cpp
```diff
@@ -51,6 +55,7 @@ const std::string& Parameter::getStringDefaultValue() const {
 
 const std::vector<std::string>& Parameter::getStringListDefaultValue() const {
     static std::vector<std::string> values;
+    values.clear();
```
mathbagu, line 58: I think this is not the good way to fix the issue: I think we should return a `vector<std::string>` instead of a reference. The `ConversionParameter::readStringListParameter` already returns a copy, maybe we can rely on the move constructor if we want to avoid two copies.

A mutex would not solve the issue, as the reference can be keep by the caller, and the value modified later.

test/iidm/converter/ConversionParametersTest.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    std::vector<std::string> stringList;
+    stringList.emplace_back("Test1Property");
+    stringList.emplace_back("Test2Property");
+    std::vector<std::string> stringList2;
+    stringList2.emplace_back("Test1Param");
+    stringList2.emplace_back("Test2Param");
+    properties.set("test-param-string-list", boost::algorithm::join(stringList, ","));
+    Parameter paramStringList("test-param-string-list", Parameter::Type::STRING_LIST, "", boost::algorithm::join(stringList2, ","));
+    Parameter paramStringList2("test-param-string-list2", Parameter::Type::STRING_LIST, "", boost::algorithm::join(stringList2, ","));
```
mathbagu, line 51: Is it more efficient to consider the values as string, instead of building them programmatically?

test/iidm/converter/ConversionParametersTest.cpp
```diff
@@ -0,0 +1,70 @@
...
+    BOOST_CHECK_EQUAL("Test1Property", ConversionParameters::readStringListParameter(properties, paramStringList)[0]);
+    BOOST_CHECK_EQUAL("Test2Property", ConversionParameters::readStringListParameter(properties, paramStringList)[1]);
+    BOOST_CHECK_EQUAL("Test1Param", ConversionParameters::readStringListParameter(properties, paramStringList2)[0]);
+    BOOST_CHECK_EQUAL("Test2Param", ConversionParameters::readStringListParameter(properties, paramStringList2)[1]);
+
+    properties.set("test-param-double", "0.06");
+    Parameter paramDouble("test-param-double", Parameter::Type::DOUBLE, "", "0.08");
+    Parameter paramDouble2("test-param-double2", Parameter::Type::DOUBLE, "", "0.08");
+    BOOST_CHECK_CLOSE(0.06, ConversionParameters::readDoubleParameter(properties, paramDouble), std::numeric_limits<double>::epsilon());
+    BOOST_CHECK_CLOSE(0.08, ConversionParameters::readDoubleParameter(properties, paramDouble2), std::numeric_limits<double>::epsilon());
```
mathbagu, line 61: Could you add a test with a `NaN`?

mathbagu (changes requested):

test/iidm/converter/ConversionParametersTest.cpp
sebalaig: Done

test/iidm/converter/ConversionParametersTest.cpp
sebalaig: You're right, it's more readable with hardcoded strings

src/iidm/converter/Parameter.cpp
sebalaig: Done

test/iidm/converter/ConversionParametersTest.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include <limits>
+#include <string>
+
+#include <boost/algorithm/string/join.hpp>
+#include <boost/test/unit_test.hpp>
+
+#include <powsybl/stdcxx/Properties.hpp>
+
+#include "iidm/converter/ConversionParameters.hpp"
+#include "iidm/converter/Parameter.hpp"
```
mathbagu, line 17: Fix these includes: these class are part of the public API

mathbagu (changes requested):

mathbagu (approved):

test/iidm/converter/ConversionParametersTest.cpp
mathbagu: ConversionParameters and Parameter are in the implementation

Issue #163, opened 2020-07-23
[IIDM 1.3] Addition of an alias system on Identifiables in IIDM


MioRtia:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
An equipment can only be identified and retrieved by its id.

* **What is the expected behavior?**
An equipment can now be identified and retrieved by one or several aliases.

* **What is the motivation / use case for changing the behavior?**
It can be useful to navigate in a network from a format with more identifiable components (e.g. CGMES) or for merging.

* **Please tell us about your environment:**
  - PowSyBl Version: 3.5.0

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

cf. powsybl/powsybl-core#1319

PR #199, opened 2020-10-06
[IIDM v1.3] Aliases on identifiables (#163)

sebalaig:

#163 

**Other information**:


src/iidm/NetworkIndex.cpp
```diff
@@ -117,12 +160,28 @@ void NetworkIndex::remove(Identifiable& identifiable) {
...
 }
 
+void NetworkIndex::removeAlias(const Identifiable& obj, const std::string& alias) {
+    auto it = m_idByAlias.find(alias);
+    if (it == m_idByAlias.end()) {
+        throw PowsyblException(stdcxx::format("No alias '%1%' found in the network", alias));
+    }
+    const std::string& idForAlias = it->second;
+    if (idForAlias != obj.getId()) {
+        throw PowsyblException(stdcxx::format("Alias '%1%' do not correspond to object '%2%'", alias, obj.getId()));
```
sebalaig, line 180: Should it be fixed to "**does** not correspond" ?

include/powsybl/iidm/Identifiable.hpp
```diff
@@ -54,12 +59,16 @@ class Identifiable : public virtual Validable, public Extendable, public MultiVa
 
     stdcxx::const_range<std::string> getPropertyNames() const;
 
+    bool hasAliases() const;
```
sebalaig, line 62: I backported this method, but unlike in Java, this method is not really useful in C++ : 
 - in Java, it is needed to know whether a **start** or an **empty** element should be written to generate an optimized XML (i.e to generate `<element />` instead of `<element></element>`)
 - in C++, the libXml2 API is quite different and both start+end elements have to be written, eventhe element have no child (i.e to generate an "empty" element in Java semantic)

So, should I remove it ?

include/powsybl/iidm/NetworkIndex.hxx
```diff
@@ -39,6 +39,9 @@ T& NetworkIndex::checkAndAdd(std::unique_ptr<T>&& identifiable) {
     }
 
     auto ptrIdentifiable = std::unique_ptr<Identifiable, Deleter>(identifiable.release());
+    for (const std::string& alias : ptrIdentifiable->getAliases()) {
```
sebalaig, line 42: Notice : this code is not really needed because aliases are added after instantiation and manually by calling `Identifiable::addAlias()`

src/iidm/NetworkIndex.cpp
```diff
@@ -70,6 +98,21 @@ Identifiable& NetworkIndex::get(const std::string& id) {
     return const_cast<Identifiable&>(static_cast<const NetworkIndex*>(this)->get<Identifiable>(id));
 }
 
+const Identifiable& NetworkIndex::get(const std::string& idOrAlias) const {
+    auto aliasIt = m_idByAlias.find(idOrAlias);
+    const std::string& id = aliasIt != m_idByAlias.end() ? aliasIt->second : idOrAlias;
```
mathbagu, line 103: You should create a utility function in `stdcxx/map.hpp` to simulate the getOrDefault java method

src/iidm/NetworkIndex.cpp
```diff
@@ -133,12 +176,28 @@ void NetworkIndex::remove(Identifiable& identifiable) {
...
         }
     }
 }
 
+void NetworkIndex::removeAlias(const Identifiable& obj, const std::string& alias) {
+    auto it = m_idByAlias.find(alias);
+    if (it == m_idByAlias.end()) {
+        throw PowsyblException(stdcxx::format("No alias '%1%' found in the network", alias));
+    }
+    const std::string& idForAlias = it->second;
```
mathbagu, line 194: This local variable is not necessary

src/iidm/converter/xml/AliasesXml.cpp
```diff
@@ -0,0 +1,47 @@
...
+namespace powsybl {
+
+namespace iidm {
+
+namespace converter {
+
+namespace xml {
+
+void AliasesXml::read(Identifiable& identifiable, const NetworkXmlReaderContext& context) {
+    identifiable.addAlias(context.getReader().readUntilEndElement(ALIAS));
```
mathbagu, line 29: I think it could be nice to have a `getElementTest()` method in XmlStreamReader (maybe the existing `getText()`). You forgot to deanonymize the alias.

src/iidm/converter/xml/NetworkXml.cpp
```diff
@@ -27,8 +27,10 @@
 
 #include "xml/XmlEncoding.hpp"
 
+#include "AliasesXml.hpp"
 #include "HvdcLineXml.hpp"
 #include "LineXml.hpp"
+#include "PropertiesXml.hpp"
```
mathbagu, line 33: It's strange you have to add it now. Fro which file does this include came from?

include/powsybl/iidm/NetworkIndex.hxx
```diff
@@ -49,14 +52,9 @@ T& NetworkIndex::checkAndAdd(std::unique_ptr<T>&& identifiable) {
 
 template<typename T>
 const T& NetworkIndex::get(const std::string& id) const {
-    checkId(id);
-
-    const auto& it = m_objectsById.find(id);
-    if (it == m_objectsById.end()) {
-        throw PowsyblException(stdcxx::format("Unable to find to the identifiable '%1%'", id));
-    }
+    const Identifiable& obj = get(id);
```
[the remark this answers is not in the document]
sebalaig: No it is 
```
    const Identifiable& get(const std::string& idOrAlias) const;
```

src/iidm/converter/xml/NetworkXml.cpp
sebalaig: I did not have to, it was probably undirectly included, I prefered to add it now even it is not relevant in this PR :/
It came from 
#include "HvdcLineXml.hpp"
#include "LineXml.hpp"

Both inherit from template class `AbstractIdentifiableXml` which needs it...

I revert ?

Issue #188, opened 2020-09-23
[IIDM v1.3] Implement SlackTerminal extension


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
There is no way to store the slack terminal of a network. **Do not integration the serializer because it's part of XIIDM v1.4** (see. https://github.com/powsybl/powsybl-core/pull/1446)

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
Add the SlackTerminal extension as done in https://github.com/powsybl/powsybl-core/pull/1381

* **What is the motivation / use case for changing the behavior?**
Have both Java and C++ implementation as closed as powsybl

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #200, opened 2020-10-13
[IIDM v1.3] Add SlackTerminal extension

sebalaig:

#188 

**Other information**:

### DIFF
--- a/include/powsybl/iidm/util/TerminalFinder.hpp
+++ b/include/powsybl/iidm/util/TerminalFinder.hpp
@@ -1,0 +1,66 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#ifndef POWSYBL_IIDM_UTIL_TERMINALFINDER_HPP
+#define POWSYBL_IIDM_UTIL_TERMINALFINDER_HPP
+
+#include <functional>
+#include <vector>
+
+#include <powsybl/stdcxx/range.hpp>
+#include <powsybl/stdcxx/reference_wrapper.hpp>
+
+namespace powsybl {
+
+namespace iidm {
+
+class Terminal;
+
+namespace util {
+
+class TerminalFinder {
+public:
+    using Comparator = std::function<bool(const Terminal& t1, const Terminal& t2)>;
+
+    using Predicate = std::function<bool(const Terminal&)>;
+
+public:
+    static TerminalFinder getDefault();
+
+public:
+    explicit TerminalFinder(const std::vector<Predicate>& predicates);
+
+    explicit TerminalFinder(const Comparator& comparator);
+
+    TerminalFinder(const TerminalFinder&) = delete;
+
+    TerminalFinder(TerminalFinder&&) noexcept = default;
+
+    TerminalFinder& operator=(const TerminalFinder&) = delete;
+
+    TerminalFinder& operator=(TerminalFinder&&) noexcept = default;
+
+    stdcxx::CReference<Terminal> find(const stdcxx::const_range<Terminal>& terminals) const;
+
+    stdcxx::Reference<Terminal> find(const stdcxx::range<Terminal>& terminals) const;
+
+private:
+    static Comparator getComparator(const std::vector<Predicate>& predicates);
+
+    static const std::vector<Predicate>& getDefaultRules();
+
+private:
+    Comparator m_comparator;
+};
+
+}  // namespace util
+
+}  // namespace iidm
+
+}  // namespace powsybl
+
+#endif  // POWSYBL_IIDM_UTIL_TERMINALFINDER_HPP
--- a/include/powsybl/network/FictitiousSwitchFactory.hpp
+++ b/include/powsybl/network/FictitiousSwitchFactory.hpp
@@ -29,9 +29,6 @@ public:
 
 public:
     FictitiousSwitchFactory() = delete;
-
-private:
-    static void createSwitch(iidm::VoltageLevel& vl, const std::string& id, const std::string& name, const iidm::SwitchKind& kind, bool retained, bool open, bool fictitious, unsigned long node1, unsigned long node2);
 };
 
 }  // namespace network
--- a/include/powsybl/network/FictitiousSwitchFactory.hpp
+++ b/include/powsybl/network/FictitiousSwitchFactory.hpp
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#ifndef POWSYBL_NETWORK_FICTITIOUSSWITCHFACTORY_HPP
+#define POWSYBL_NETWORK_FICTITIOUSSWITCHFACTORY_HPP
+
+#include <string>
+
+#include <powsybl/iidm/Network.hpp>
+#include <powsybl/iidm/SwitchKind.hpp>
+
+namespace powsybl {
+
+namespace iidm {
+
+class VoltageLevel;
+
+}  // namespace iidm
+
+namespace network {
+
+class FictitiousSwitchFactory {
+public:
+    static iidm::Network create();
+
+public:
+    FictitiousSwitchFactory() = delete;
+
+private:
+    static void createSwitch(iidm::VoltageLevel& vl, const std::string& id, const std::string& name, const iidm::SwitchKind& kind, bool retained, bool open, bool fictitious, unsigned long node1, unsigned long node2);
+};
+
+}  // namespace network
+
+}  // namespace powsybl
+
+#endif  // POWSYBL_NETWORK_FICTITIOUSSWITCHFACTORY_HPP
--- a/include/powsybl/stdcxx/instanceof.hpp
+++ b/include/powsybl/stdcxx/instanceof.hpp
@@ -26,12 +26,12 @@ inline bool isInstanceOf(const std::unique_ptr<T>& pointer) {
 
 template <typename Base, typename T>
 inline bool isInstanceOf(const stdcxx::CReference<T>& object) {
-    return static_cast<bool>(object) && std::is_same<Base, T>::value ? true : dynamic_cast<const Base*>(&object.get()) != nullptr;
+    return static_cast<bool>(object) && isInstanceOf<Base>(object.get());
 }
 
 template <typename Base, typename T>
 inline bool isInstanceOf(const stdcxx::Reference<T>& object) {
-    return static_cast<bool>(object) && std::is_same<Base, T>::value ? true : dynamic_cast<const Base*>(&object.get()) != nullptr;
+    return static_cast<bool>(object) && isInstanceOf<Base>(object.get());
 }
 
 }  // namespace stdcxx
--- a/include/powsybl/stdcxx/instanceof.hpp
+++ b/include/powsybl/stdcxx/instanceof.hpp
@@ -10,6 +10,8 @@
 
 #include <memory>
 
+#include <powsybl/stdcxx/reference_wrapper.hpp>
+
 namespace stdcxx {
 
 template <typename Base, typename T>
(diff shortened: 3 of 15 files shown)


include/powsybl/iidm/util/TerminalFinder.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+namespace iidm {
+
+class Terminal;
+
+namespace util {
+
+class TerminalFinder {
+public:
+    using Comparator = std::function<bool(Terminal& t1, Terminal& t2)>;
```
mathbagu, line 27: Could the terminals be const?

include/powsybl/iidm/util/TerminalFinder.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+    using Predicate = std::function<bool(Terminal&)>;
```
mathbagu, line 29: I think the terminal should be declared const

include/powsybl/iidm/util/TerminalFinder.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+public:
+    static TerminalFinder getDefault();
+
+    static const std::vector<Predicate>& getDefaultRules();
+
+public:
+    explicit TerminalFinder(const std::vector<Predicate>& predicates);
```
mathbagu, line 37: In java, there is also a constructor that take a comparator

include/powsybl/iidm/util/TerminalFinder.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+    stdcxx::Reference<Terminal> find(const stdcxx::range<Terminal>& terminals) const;
```
mathbagu, line 39: I wonder if we should provide a const version of this method

include/powsybl/iidm/util/TerminalFinder.hpp
```diff
@@ -0,0 +1,54 @@
...
+class TerminalFinder {
+public:
+    using Comparator = std::function<bool(Terminal& t1, Terminal& t2)>;
+
+    using Predicate = std::function<bool(Terminal&)>;
+
+public:
+    static TerminalFinder getDefault();
+
+    static const std::vector<Predicate>& getDefaultRules();
```
mathbagu, line 34: In java, this method is declared private

include/powsybl/network/FictitiousSwitchFactory.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+class FictitiousSwitchFactory {
+public:
+    static iidm::Network create();
+
+private:
+    static void createSwitch(iidm::VoltageLevel& vl, const std::string& id, const std::string& name, const iidm::SwitchKind& kind, bool retained, bool open, bool fictitious, unsigned long node1, unsigned long node2);
+
+public:
+    FictitiousSwitchFactory() = delete;
```
mathbagu, line 34: The public method should be declare before private method, even if they are static

src/iidm/extensions/SlackTerminal.cpp
```diff
@@ -0,0 +1,122 @@
...
+    VoltageLevel& vl = bus.getVoltageLevel();
+    stdcxx::Reference<Terminal> terminal = util::TerminalFinder::getDefault().find(bus.getConnectedTerminals());
+    if (!terminal) {
+        throw PowsyblException(stdcxx::format("Unable to find a terminal in the bus %1%", bus.getId()));
+    }
+    reset(vl, terminal);
+}
+
+void SlackTerminal::deleteVariantArrayElement(unsigned long index) {
+    m_terminals[index] = stdcxx::ref<Terminal>();
```
mathbagu, line 52: Use `m_terminals[index].reset()` instead?

src/iidm/extensions/SlackTerminal.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+const std::type_index& SlackTerminal::getType() const {
+    static std::type_index s_type = typeid(SlackTerminal);
+    return s_type;
+}
+
+bool SlackTerminal::isCleanable() const {
+    return std::find_if(m_terminals.begin(), m_terminals.end(), [](const stdcxx::Reference<Terminal>& terminal) {
+        return static_cast<bool>(terminal);
+    }) == m_terminals.end();
```
mathbagu, line 80: I would prefer you store the result of `find_if` in a local variable to increase the readability

src/iidm/util/TerminalFinder.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+TerminalFinder::Comparator TerminalFinder::getComparator(const std::vector<Predicate>& predicates) {
+    return [&predicates](Terminal& t1, Terminal& t2) {
+        if (stdcxx::areSame(t1, t2)) {
+            return false;
+        }
+        for (const auto& predicate : predicates) {
+            if (predicate(t1) || predicate(t2)) {
+                return !predicate(t1) && predicate(t2);
+            }
```
mathbagu, line 40: This code is more complex to understand than the java version. The comments have been removed and I'm not this is equivalent. Is there a good reason to write it like this?

src/iidm/util/TerminalFinder.cpp
```diff
@@ -0,0 +1,65 @@
...
+    return TerminalFinder(getDefaultRules());
+}
+
+const std::vector<TerminalFinder::Predicate>& TerminalFinder::getDefaultRules() {
+    static std::vector<Predicate> s_rules = {
+        [](Terminal& t) { return t.getConnectable() && stdcxx::isInstanceOf<BusbarSection>(t.getConnectable().get()); },
+        [](Terminal& t) { return t.getConnectable() && stdcxx::isInstanceOf<Injection>(t.getConnectable().get()); },
+        [](Terminal& t) { return t.getConnectable() && stdcxx::isInstanceOf<Branch>(t.getConnectable().get()); },
+        [](Terminal& t) { return t.getConnectable() && stdcxx::isInstanceOf<ThreeWindingsTransformer>(t.getConnectable().get()); },
+        [](Terminal& t) { return t.getConnectable() && stdcxx::isInstanceOf<HvdcConverterStation>(t.getConnectable().get()); }
```
mathbagu, line 56: It would be nice to create a new version of the `stdcxx::isInstanceOf` that takes a reference_wrapper<T>.

test/iidm/extensions/SlackTerminalTest.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    POWSYBL_ASSERT_THROW(adder.add(), PowsyblException, "Terminal needs to be set to create a SlackTerminal extension");
+
+    // Defining the slackTerminal on the first Terminal
+    const std::string& busBusBreakerId = "B";
+    Terminal& terminal = *boost::begin(network.getBusBreakerView().getBus(busBusBreakerId).get().getConnectedTerminals());
+    adder.withTerminal(terminal).add();
+
+    for (VoltageLevel& vl : network.getVoltageLevels()) {
+        auto slackTerminal = vl.findExtension<SlackTerminal>();
```
mathbagu, line 110: const auto& ?

test/iidm/extensions/SlackTerminalTest.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    // Adding a terminal in the wrong voltage level
+    Terminal& wrongTerminal = *network.getBusBreakerView().getBus("NLOAD").get().getConnectedTerminals().begin();
+    SlackTerminalAdder slackTerminalAdder = vl.newExtension<SlackTerminalAdder>().withTerminal(wrongTerminal);
+    POWSYBL_ASSERT_THROW(slackTerminalAdder.add(), PowsyblException, "Terminal given is not in the right VoltageLevel (VLLOAD instead of VLHV1)");
+
+    // First adding a terminal in the right voltage level...
+    Terminal& terminal = *boost::begin(network.getBusBreakerView().getBus("NHV1").get().getConnectedTerminals());
+    slackTerminalAdder.withTerminal(terminal).add();
+    SlackTerminal slackTerminal = vl.getExtension<SlackTerminal>();
```
mathbagu, line 163: Extension should not be copiable?

test/iidm/util/TerminalFinderTest.cpp
```diff
@@ -0,0 +1,108 @@
...
+    BOOST_CHECK_EQUAL(ConnectableType::BUSBAR_SECTION, bestTerminal.getConnectable().get().getType());
+    BOOST_CHECK_EQUAL("N_0", bestTerminal.getBusBreakerView().getBus().get().getId());
+    BOOST_CHECK_EQUAL("N_0", bestTerminal.getBusView().getBus().get().getId());
+}
+
+BOOST_AUTO_TEST_CASE(testNoTerminal) {
+    Network network = powsybl::network::FictitiousSwitchFactory::create();
+    VoltageLevel& vlN = network.getVoltageLevel("N");
+
+    util::TerminalFinder slackTerminalFinder = util::TerminalFinder::getDefault();
```
mathbagu, line 46: Not sure TerminalFinder should be copiable too

mathbagu (changes requested):

src/iidm/util/TerminalFinder.cpp
sebalaig: Agree with you! Actually, the problem is that a C++ comparator returns a boolean, unlike in Java where it may return an (negative, 0 or positive) integer. 

In C++, returning -1 is equivalent to true, that's why Java implementation could not be copied/paste here.

I copied/pasted Java test suite (see `TerminalFinderTest.cpp`), so that we check that we stricltly have the same behaviour.

test/iidm/util/TerminalFinderTest.cpp
sebalaig: Why ? The comparator is the only attribute and is copiable.

test/iidm/extensions/SlackTerminalTest.cpp
sebalaig: This extension is copiable like every other extension. Do you mean by deleting copy constructor and affectation operator ? If yes, to be done in a separated PR (IMHO).

include/powsybl/iidm/util/TerminalFinder.hpp
sebalaig: Yes they should be => updated

test/iidm/extensions/SlackTerminalTest.cpp
mathbagu: But this is a mistake: extensions should not be copiable because there is a double-link between the extendable and the extension. By definition, an extendable can carry only a single instance of each extensions.

Agree to do it in a separate PR: could you please prepare an issue and assign it in the current project. From my point of view, this is a critical design issue

test/iidm/util/TerminalFinderTest.cpp
mathbagu: Because this can be considered as a singleton that is not copyable by definition

test/iidm/extensions/SlackTerminalTest.cpp
sebalaig: Ok, done : https://github.com/powsybl/powsybl-iidm4cpp/issues/203

test/iidm/util/TerminalFinderTest.cpp
sebalaig: Ok

include/powsybl/network/FictitiousSwitchFactory.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+class FictitiousSwitchFactory {
+public:
+    static iidm::Network create();
+
+public:
+    FictitiousSwitchFactory() = delete;
+
+private:
+    static void createSwitch(iidm::VoltageLevel& vl, const std::string& id, const std::string& name, const iidm::SwitchKind& kind, bool retained, bool open, bool fictitious, unsigned long node1, unsigned long node2);
```
mathbagu, line 34: this could be a function in the c++ file

include/powsybl/stdcxx/instanceof.hpp
```diff
@@ -22,6 +24,16 @@ inline bool isInstanceOf(const std::unique_ptr<T>& pointer) {
     return dynamic_cast<Base*>(pointer.get()) != nullptr;
 }
 
+template <typename Base, typename T>
+inline bool isInstanceOf(const stdcxx::CReference<T>& object) {
+    return static_cast<bool>(object) && std::is_same<Base, T>::value ? true : dynamic_cast<const Base*>(&object.get()) != nullptr;
```
mathbagu, line 29: why don't you call the `isInstance(const T&)` functions instead duplicate the code?

include/powsybl/stdcxx/instanceof.hpp
```diff
@@ -22,6 +24,16 @@ inline bool isInstanceOf(const std::unique_ptr<T>& pointer) {
...
+}
+
+template <typename Base, typename T>
+inline bool isInstanceOf(const stdcxx::Reference<T>& object) {
+    return static_cast<bool>(object) && std::is_same<Base, T>::value ? true : dynamic_cast<const Base*>(&object.get()) != nullptr;
```
mathbagu, line 34: Same question?

src/iidm/extensions/SlackTerminalAdder.cpp
```diff
@@ -0,0 +1,52 @@
...
+SlackTerminalAdder::SlackTerminalAdder(Extendable& extendable) :
+    ExtensionAdder(extendable) {
+}
+
+std::unique_ptr<Extension> SlackTerminalAdder::createExtension(Extendable& extendable) {
+    if (!m_terminal) {
+        throw PowsyblException("Terminal needs to be set to create a SlackTerminal extension");
+    }
+    if (stdcxx::isInstanceOf<VoltageLevel>(extendable)) {
+        const auto& voltageLevel = dynamic_cast<const VoltageLevel&>(extendable);
```
mathbagu, line 29: remove the `const` qualifier and use `voltageLevel` at line 33?

mathbagu (approved):

PR #211, opened 2020-11-04
Fix multi-variant implementation by adding the Network in the index

mathbagu:

N/A

Bug fix

Currently the network is not in the index. When a variant is created or removed, the network is not notified. If a multi-variant extension is attached to the network, the extension fields are not extended.

This PR insert the network in its own index. To do that, we need to customize the `Deleter` parameter of `std::unique_ptr` to avoid to delete the network twice.

**Other information**:

### DIFF
--- a/include/powsybl/stdcxx/instanceof.hpp
+++ b/include/powsybl/stdcxx/instanceof.hpp
@@ -14,7 +14,12 @@ namespace stdcxx {
 
 template <typename Base, typename T>
 inline bool isInstanceOf(const T& object) {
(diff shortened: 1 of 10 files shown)


include/powsybl/stdcxx/instanceof.hpp
```diff
@@ -12,9 +12,14 @@
 
 namespace stdcxx {
 
+template <typename Base, typename T>
+inline bool isInstanceOf(const T* object) {
+    return object != nullptr && dynamic_cast<const Base*>(object) != nullptr;
```
sebalaig, line 17: Not sure the check `object != nullptr` is necessary

Issue #191, opened 2020-09-23
[XIIDM v1.3] Implement BusbarSectionPosition and ConnectablePosition


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
These extensions already exist in a private repository. They have removed because they are duplicated in the powsybl-single-line-diagram repository.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
Have these extensions in a new **sld** module.

* **What is the motivation / use case for changing the behavior?**
Have both implementation as closed as possible

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #214, opened 2020-11-13
[IIDM v1.3] Add BusbarSectionPosition (#191)

sebalaig:

#191 

**Other information**:


extensions/sld/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
+set(EXT_NAME ext-sld)
+
+set(EXT_SOURCES
+    src/BusbarSectionPosition.cpp
+    src/BusbarSectionPositionAdder.cpp
+    src/BusbarSectionPositionXmlSerializer.cpp
+    src/Sld.cpp
```
mathbagu, line 14: I wonder if we should rename this file sld.cpp as it doesn't contain any types. It's done like this for other modules, so maybe we can do this in another PR.

extensions/sld/src/BusbarSectionPositionAdder.cpp
```diff
@@ -0,0 +1,56 @@
...
+    ExtensionAdder(extendable) {
+}
+
+std::unique_ptr<Extension> BusbarSectionPositionAdder::createExtension(Extendable& extendable) {
+    if (!m_busbarIndex.is_initialized()) {
+        throw PowsyblException("Undefined value for busbar index");
+    }
+    if (!m_sectionIndex) {
+        throw PowsyblException("Undefined value for section index");
+    }
```
mathbagu, line 33: Choose the way you want to check indexes are set, but use the same syntax wherever. It seems we are not consistent in the API. Could you also create an issue to fix these inconsistencies?

mathbagu (changes requested):

mathbagu (approved):

PR #215, opened 2020-11-13
[IIDM v1.3] Add ConnectablePosition extension (#191)

sebalaig:

#191 

**Other information**:

### DIFF
--- a/extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
+++ b/extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
@@ -62,7 +62,7 @@ public:  // Extension
     const std::type_index& getType() const override;
 
 public:
-    ConnectablePosition(Connectable& connectable, OptionalFeeder&& feeder, OptionalFeeder&& feeder1, OptionalFeeder&& feeder2, OptionalFeeder&& feeder3);
+    ConnectablePosition(Connectable& connectable, const OptionalFeeder& feeder, const OptionalFeeder& feeder1, const OptionalFeeder& feeder2, const OptionalFeeder& feeder3);
 
     const OptionalFeeder& getFeeder() const;
 
--- a/extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
+++ b/extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
@@ -62,7 +62,7 @@ public:  // Extension
     const std::type_index& getType() const override;
 
 public:
-    ConnectablePosition(Connectable& connectable, const OptionalFeeder& feeder, const OptionalFeeder& feeder1, const OptionalFeeder& feeder2, const OptionalFeeder& feeder3);
+    ConnectablePosition(Connectable& connectable, OptionalFeeder&& feeder, OptionalFeeder&& feeder1, OptionalFeeder&& feeder2, OptionalFeeder&& feeder3);
 
     const OptionalFeeder& getFeeder() const;
 
@@ -79,8 +79,6 @@ public:
     const OptionalFeeder& getFeeder3() const;
 
     OptionalFeeder& getFeeder3();
-
-    ConnectablePosition& setFeeders(const OptionalFeeder& feeder, const OptionalFeeder& feeder1, const OptionalFeeder& feeder2, const OptionalFeeder& feeder3);
 
 private:  // Extension
     void assertExtendable(const stdcxx::Reference<Extendable>& extendable) const override;
--- a/extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
+++ b/extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
@@ -95,8 +95,6 @@ private:
     OptionalFeeder m_feeder3;
 };
 
-std::ostream& operator<<(std::ostream& stream, const ConnectablePosition::Direction& value);
-
 }  // namespace sld
 
 }  // namespace extensions
--- a/extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
+++ b/extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
@@ -1,0 +1,101 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#ifndef POWSYBL_IIDM_EXTENSIONS_SLD_CONNECTABLEPOSITION_HPP
+#define POWSYBL_IIDM_EXTENSIONS_SLD_CONNECTABLEPOSITION_HPP
+
+#include <boost/optional.hpp>
+
+#include <powsybl/iidm/Extension.hpp>
+
+namespace powsybl {
+
+namespace iidm {
+
+class Connectable;
+
+namespace extensions {
+
+namespace sld {
+
+class ConnectablePosition : public Extension {
+public:
+    enum class Direction : unsigned char {
+        TOP,
+        BOTTOM,
+        UNDEFINED
+    };
+
+    class Feeder {
+    public:
+        Feeder(const std::string& name, unsigned long order, const Direction& direction);
+
+        const Direction& getDirection() const;
+
+        const std::string& getName() const;
+
+        unsigned long getOrder() const;
+
+        Feeder& setDirection(const Direction& direction);
+
+        Feeder& setName(const std::string& name);
+
+        Feeder& setOrder(int order);
+
+    private:
+        std::string m_name;
+
+        unsigned long m_order;
+
+        Direction m_direction;
+    };
+
+    using OptionalFeeder = boost::optional<Feeder>;
+
+public:  // Extension
+    const std::string& getName() const override;
+
+    const std::type_index& getType() const override;
+
+public:
+    ConnectablePosition(Connectable& connectable, const OptionalFeeder& feeder, const OptionalFeeder& feeder1, const OptionalFeeder& feeder2, const OptionalFeeder& feeder3);
+
+    const OptionalFeeder& getFeeder() const;
+
+    const OptionalFeeder& getFeeder1() const;
+
+    const OptionalFeeder& getFeeder2() const;
+
+    const OptionalFeeder& getFeeder3() const;
+
+    ConnectablePosition& setFeeders(const OptionalFeeder& feeder, const OptionalFeeder& feeder1, const OptionalFeeder& feeder2, const OptionalFeeder& feeder3);
+
+private:  // Extension
+    void assertExtendable(const stdcxx::Reference<Extendable>& extendable) const override;
+
+private:
+    static void check(const OptionalFeeder& feeder, const OptionalFeeder& feeder1, const OptionalFeeder& feeder2, const OptionalFeeder& feeder3);
+
+private:
+    OptionalFeeder m_feeder;
+
+    OptionalFeeder m_feeder1;
+
+    OptionalFeeder m_feeder2;
+
+    OptionalFeeder m_feeder3;
+};
+
+}  // namespace sld
+
+}  // namespace extensions
+
+}  // namespace iidm
+
+}  // namespace powsybl
+
+#endif  // POWSYBL_IIDM_EXTENSIONS_SLD_CONNECTABLEPOSITION_HPP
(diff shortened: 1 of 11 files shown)


extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePositionAdder.hpp
```diff
@@ -0,0 +1,116 @@
...
+        FeederAdder& withOrder(unsigned long order);
+
+    private:
+        std::string m_name;
+
+        stdcxx::optional<unsigned long> m_order;
+
+        stdcxx::optional<ConnectablePosition::Direction> m_direction;
+
+        std::reference_wrapper<ConnectablePositionAdder> m_parent;
```
mathbagu, line 44: Why don't you use a simple `ConnectablePositionAdder&` reference?

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePositionAdder.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+        std::reference_wrapper<ConnectablePosition::OptionalFeeder> m_feeder;
```
mathbagu, line 46: This is strange and not compatible with the expected design?

extensions/sld/src/ConnectablePosition.cpp
```diff
@@ -0,0 +1,147 @@
...
+    return m_feeder1;
+}
+
+const ConnectablePosition::OptionalFeeder& ConnectablePosition::getFeeder2() const {
+    return m_feeder2;
+}
+
+const ConnectablePosition::OptionalFeeder& ConnectablePosition::getFeeder3() const {
+    return m_feeder3;
+}
```
mathbagu, line 120: To be check: do we need also the non const version to be able to modify the feeder

extensions/sld/src/ConnectablePositionAdder.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace iidm {
+
+namespace extensions {
+
+namespace sld {
+
+ConnectablePositionAdder::ConnectablePositionAdder(Extendable& extendable) :
+    ExtensionAdder(extendable) {
+}
```
mathbagu, line 23: This should be declare after internal class methods

extensions/sld/src/ConnectablePositionAdder.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+ConnectablePositionAdder& ConnectablePositionAdder::FeederAdder::add() {
+    if (!m_order.is_initialized()) {
+        throw PowsyblException("Feeder order is not set");
+    }
+    if (!m_direction.is_initialized()) {
+        throw PowsyblException("Feeder direction is not set");
+    }
+    m_feeder.get() = ConnectablePosition::Feeder(m_name, *m_order, *m_direction);
+    return m_parent.get();
```
mathbagu, line 38: Does it work?
```
m_feeder = ConnectablePosition::Feeder(m_name, *m_order, *m_direction);
```

I think you should also check if the name is not empty. Looking to the java code, we only check if the name is null, but an empty string is weird

extensions/sld/src/ConnectablePositionXmlSerializer.cpp
```diff
@@ -0,0 +1,98 @@
...
+        } else {
+            throw AssertionError(stdcxx::format("Unexpected element: %1%", context.getReader().getLocalName()));
+        }
+    });
+    adder.add();
+    return extendable.getExtension<ConnectablePosition>();
+}
+
+void ConnectablePositionXmlSerializer::readPosition(const converter::xml::NetworkXmlReaderContext& context, ConnectablePositionAdder::FeederAdder& feederAdder) const {
+    const std::string& name = context.getReader().getAttributeValue("name");
```
mathbagu, line 58: Use `const auto&` here too ?

extensions/sld/src/ConnectablePositionAdder.cpp
sebalaig: >Does it work?
To be checked. It looks like this code should be written (as suggested [here](https://github.com/powsybl/powsybl-iidm4cpp/pull/215#discussion_r525250627) to respect Java implementation to avoid memory corruption...

> I think you should also check if the name is not empty. Looking to the java code, we only check if the name is null, but an empty string is weird
I agree, I wanted to be stricly Java compliant. => check added

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePositionAdder.hpp
sebalaig: It was in order to make FeederAdder assignable. For now on, ignore this feature because it makes implementation simpler.

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePositionAdder.hpp
sebalaig: Non compatible with the expected design => Agree
Strange : to be discussed :). In Java implementation, it uses a template specialization on-the-fly. Not sure it could be done that simple in C++.

extensions/sld/src/ConnectablePosition.cpp
sebalaig: Added

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
```diff
@@ -90,6 +98,8 @@ class ConnectablePosition : public Extension {
     OptionalFeeder m_feeder3;
 };
 
+std::ostream& operator<<(std::ostream& stream, const ConnectablePosition::Direction& value);
```
mathbagu, line 99: To be check: is it really necessary? In `include/powsybl/iidm/Enum.hpp` we have an implementation that should work with all enums. You just have to specialize the `toString()` template method

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
```diff
@@ -0,0 +1,109 @@
...
+    OptionalFeeder m_feeder;
+
+    OptionalFeeder m_feeder1;
+
+    OptionalFeeder m_feeder2;
+
+    OptionalFeeder m_feeder3;
+};
+
+std::ostream& operator<<(std::ostream& stream, const ConnectablePosition::Direction& value);
```
mathbagu, line 99: To be check: is it really necessary? In include/powsybl/iidm/Enum.hpp we have an implementation that should work with all enums. You just have to specialize the `getNames()` template method

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+    private:
+        std::string m_name;
+
+        unsigned long m_order;
+
+        Direction m_direction;
+    };
+
+    using OptionalFeeder = boost::optional<Feeder>;
```
mathbagu, line 57: To be discussed: in the other classes, we use the `stdcxx::reference_wrapper` that allow empty references. Maybe we should use the same design in the `ConnectablePosition` class, but an `OptionalFeeder` in the adder?

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePositionAdder.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Destructor
+     */
+    ~ConnectablePositionAdder() noexcept override = default;
+
+    /**
+     * Copy constructor
+     */
+    ConnectablePositionAdder(const ConnectablePositionAdder&) = default;
```
mathbagu, line 63: Add a variable name

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePositionAdder.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Move constructor
+     */
+    ConnectablePositionAdder(ConnectablePositionAdder&&) = default;
```
mathbagu, line 68: Add a variable name

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePositionAdder.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Copy assignment operator
+     */
+    ConnectablePositionAdder& operator=(const ConnectablePositionAdder&) = delete;
```
mathbagu, line 73: Add a variable name

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePositionAdder.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Move assignment operator
+     */
+    ConnectablePositionAdder& operator=(ConnectablePositionAdder&&) = delete;
```
mathbagu, line 78: Add a variable name

extensions/sld/src/ConnectablePosition.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+const std::type_index& ConnectablePosition::getType() const {
+    static std::type_index s_type = typeid(ConnectablePosition);
+    return s_type;
+}
+
+std::ostream& operator<<(std::ostream& stream, const ConnectablePosition::Direction& value) {
+    stream << iidm::Enum::toString(value);
+    return stream;
+}
```
mathbagu, line 151: Remove this function: it should not be needed

extensions/sld/src/ConnectablePosition.cpp
```diff
@@ -0,0 +1,159 @@
...
+        if (feeder2 || feeder3) {
+            error = true;
+        }
+    }
+    if (error) {
+        throw PowsyblException(stdcxx::format("feeder 1|2|3 have to be set in the right order"));
+    }
+}
+
+const ConnectablePosition::OptionalFeeder& ConnectablePosition::getFeeder() const {
```
mathbagu, line 106: To be check: do we return a `stdcxx::reference_wrapper` instead

extensions/sld/src/ConnectablePositionAdder.cpp
```diff
@@ -0,0 +1,88 @@
...
+    if (m_name.empty()) {
+        throw PowsyblException("Feeder name is empty");
+    }
+    if (!m_order.is_initialized()) {
+        throw PowsyblException("Feeder order is not set");
+    }
+    if (!m_direction.is_initialized()) {
+        throw PowsyblException("Feeder direction is not set");
+    }
+    m_feeder = ConnectablePosition::Feeder(m_name, *m_order, *m_direction);
```
mathbagu, line 36: That's really better written like this :+1:

mathbagu (changes requested):

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
sebalaig: Yes it looks it is necessary unless we cannot write (ConnectablePositionTest.cpp:80)
```
    BOOST_CHECK_EQUAL(ConnectablePosition::Direction::BOTTOM, extension.getFeeder().get().getDirection());
```

It looks like this operator should be defined in the namespace of the class to be ok, which is not the case of the generic operator defined in Enum.hpp (i.e defined in `powsybl::iidm`)

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
sebalaig: I know what you mean, but the problem with usage of reference_wrapper is about ownership of the feeder. 
However, I agree that getters should return stdcxx::reference_wrapper.

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePositionAdder.hpp
sebalaig: When defaulted/deleted, we do not write any parameter usually.

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePosition.hpp
mathbagu: The attributes are declared as `OptionalFeeder` but the getters should return a `stdcxx::Reference` or `stdcxx::CReference`

extensions/sld/include/powsybl/iidm/extensions/sld/ConnectablePositionAdder.hpp
mathbagu: Create an issue to remove the variable name for defaulted/deleted constructors and assignments.

Add a statement in the `wiki/coding-style` to explain that we do not give the variable name for defaulted/deleted constructors or assignment operators

mathbagu (approved):

Issue #49, opened 2020-03-13
[IIDM v1.0] Implement the connected/synchronous connected component


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
The network is not able to compute the connected/synchronous components.
The [PR#1445](https://github.com/powsybl/powsybl-core/pull/1445) aims to refactor the feature, so this is the good time to add this feature to the C++ implementation

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
To have a convenient method to get from a bus it's connected/synchronous component number.

* **What is the motivation / use case for changing the behavior?**
Have both Java and C++ API as close as possible

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

mathbagu:
@thiebarrrte starts working on in in the `add_components` branch. It's almost finish: we pause the development due to a potential bug in the Java implementation.

mathbagu:
@tadam50 is working on the refactoring of the components in the Java implementation. I propose to wait he finishes before continuing this PR. Maybe we can move this to IIDM v1.1

PR #217, opened 2020-11-20
Add connected/synchronous components

mathbagu:

#49

Feature

The connected and synchronous components are not implemented

The connected and synchronous components are now available

**Other information**:


include/powsybl/iidm/Network.hpp
```diff
@@ -161,6 +163,10 @@ class Network : public Container, public VariantManagerHolder {
     template <typename T = Connectable, typename = typename std::enable_if<std::is_base_of<Connectable, T>::value>::type>
     stdcxx::range<T> getConnectables();
 
+    const ConnectedComponentsManager& getConnectedComponentsManager() const;
+
+    ConnectedComponentsManager& getConnectedComponentsManager();
+
```
sebalaig, line 169: These methods should be private as they are not in Java api

include/powsybl/iidm/Network.hpp
```diff
@@ -289,6 +295,10 @@ class Network : public Container, public VariantManagerHolder {
 
     stdcxx::range<Switch> getSwitches();
 
+    const SynchronousComponentsManager& getSynchronousComponentsManager() const;
+
+    SynchronousComponentsManager& getSynchronousComponentsManager();
+
```
sebalaig, line 301: Same remark

include/powsybl/iidm/util/ConnectedComponents.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include <powsybl/stdcxx/optional.hpp>
+#include <powsybl/stdcxx/reference_wrapper.hpp>
+
+namespace powsybl {
+
+namespace iidm {
+
+class Bus;
+
+namespace ConnectedComponents {
```
sebalaig, line 20: This is not related to this PR, but classes/methods/functions in the "util" directory does not always are in namespace `powsybl::iidm::util`. I think either all or none of them should be.

include/powsybl/math/GraphUtil.hpp
```diff
@@ -0,0 +1,44 @@
...
+    using ConnectedComponentsComputationResult = graph_util::ConnectedComponentsComputationResult;
+
+public:
+    static ConnectedComponentsComputationResult computeConnectedComponents(const std::vector<std::vector<unsigned long>>& adjacencyList);
+
+public:
+    GraphUtil() = delete;
+
+    template <typename V, typename E>
+    void removeIsolatedVertices(UndirectedGraph<V, E>& graph);
```
sebalaig, line 32: What is this method about ? It should be removed (not called, and GraphUtil is not instanciable).

include/powsybl/math/GraphUtil.hxx
```diff
@@ -0,0 +1,36 @@
...
+#define POWSYBL_MATH_GRAPHUTIL_HXX
+
+#include <powsybl/math/GraphUtil.hpp>
+
+namespace powsybl {
+
+namespace math {
+
+template <typename V, typename E>
+void GraphUtil::removeIsolatedVertices(UndirectedGraph<V, E>& graph) {
```
sebalaig, line 18: To be removed

src/iidm/NetworkVariant.cpp
```diff
@@ -16,17 +18,37 @@ namespace iidm {
...
-VariantImpl::VariantImpl(Network& network, VariantImpl&& /*variant*/) noexcept :
-    Variant(network) {
+VariantImpl::VariantImpl(Network& network, VariantImpl&& variant) noexcept :
+    Variant(network),
+    m_connectedComponentsManager(network, std::move(variant.m_connectedComponentsManager)),
+    m_synchronousComponentsManager(network, std::move(variant.m_synchronousComponentsManager)) {
 }
 
 std::unique_ptr<VariantImpl> VariantImpl::copy() const {
     return stdcxx::make_unique<VariantImpl>(m_owner);
```
sebalaig, line 33: To be discussed : I know that components managers will recompute components if needed later, but the fact that `copy` does not copy components managers may be a bit disturbing.

include/powsybl/math/ConnectedComponentsComputationResult.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace powsybl {
+
+namespace math {
+
+namespace graph_util {
+
+class ConnectedComponentsComputationResult {
+public:
+    ConnectedComponentsComputationResult(std::vector<stdcxx::optional<unsigned long>>&& componentNumber, const std::vector<unsigned long>& orderedComponents);
```
sebalaig, line 23: `orderedComponents` could be moved also.

sebalaig (changes requested):
Probably that we can update `BusBreakerVoltageLevel::invalidateCache()` and uncomment the following lines : 
```
    // getNetwork().getConnectedComponentsManager().invalidate();
    // getNetwork().getSynchronousComponentsManager().invalidate();
```

include/powsybl/iidm/util/ConnectedComponents.hpp
mathbagu: https://github.com/powsybl/powsybl-iidm4cpp/issues/222

include/powsybl/math/GraphUtil.hpp
mathbagu: Looking to the Java API, this method is static. This is an utility class, and according to #222, we should switch to a namespace also here.

include/powsybl/math/GraphUtil.hxx
mathbagu: This method is used in Java, so we probably miss something. If it's not called, we probably miss a change in the Node/Breaker VL implementation.

include/powsybl/math/GraphUtil.hpp
mathbagu: This method is defined in UndirectedGraph, so we can safely remove it

src/iidm/NetworkVariant.cpp
mathbagu: Same in java, so I propose to keep like this even if I agree with you

mathbagu: https://github.com/powsybl/powsybl-core/issues/1548

include/powsybl/iidm/Network.hpp
mathbagu: It's true, but this method is needed by a lot of classes in the implementation, so I have to declare lot of classes as friend. I'm not sure it's better. The ComponentManager provides only one risky method which is the `invalidate()`. This will lead to a complete calculation of the connectivity.

It seems acceptable to keep these methods public.

sebalaig (approved):

PR #231, opened 2020-12-11
Fix danglingline adder

mathbagu:

https://github.com/powsybl/powsybl-core/pull/1568

Refactoring

The generation is set after the DanglingLine creation

The generation is pass to the DanglingLine constructor, and then attached to the danglingLine.

**Other information**:


src/iidm/DanglingLine.cpp
```diff
@@ -15,15 +15,21 @@ namespace powsybl {
 namespace iidm {
 
 DanglingLine::DanglingLine(VariantManagerHolder& network, const std::string& id, const std::string& name, bool fictitious,
-                           double p0, double q0, double r, double x, double g, double b, const std::string& ucteXnodeCode) :
+                           double p0, double q0, double r, double x, double g, double b, const std::string& ucteXnodeCode,
+                           std::unique_ptr<Generation>&& generation) :
```
sebalaig, line 19: Probably that storing the generation in an optional rather than a unique_ptr should be simpler.

include/powsybl/stdcxx/reference_wrapper.hpp
```diff
@@ -33,6 +33,11 @@ class reference_wrapper {
 
     reference_wrapper(reference_wrapper&&) noexcept = default;
 
+    reference_wrapper& operator=(T& reference) {
```
sebalaig, line 36: A const version may be usefull also, what do you think ?

sebalaig (changes requested):

include/powsybl/stdcxx/reference_wrapper.hpp
mathbagu: The `const` qualifier is carried by the T. You should not be able to assign a `const T&` to a `Reference<T>`. This should be possible only with `CReference<T>`. This is the reason why the T is not const.

src/iidm/DanglingLine.cpp
mathbagu: I would try to update the implementation with an optional. I would also try it for the ShuntCompensatorModel.

If it works well, I will create an issue to change all our unique_ptr usage by optional

sebalaig (approved):

PR #254, opened 2020-12-22
Add missing alias resolution when calling NetworkIndex::find()

sebalaig:

No

Bug fix

`Network::find` does not consider aliases at all (unlike `Network::get`)

Network::find tries to resolve alias before searching in the id map

**Other information**:


mathbagu (commented):
To avoid duplication, I think we should refactor the `get<>` method to call `find` and handle empty references. What do you think?

Issue #210, opened 2020-11-04
Allow to get an extension by its name


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
It's possible to get an extension by it's type, but it's not possible to get it by its name. This feature exists in the java implementation.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
Be able to get an extension by its name.

* **What is the motivation / use case for changing the behavior?**
Have the same implementation in Java and C++

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

sebalaig:
Closed by #259

PR #259, opened 2021-01-07
[IIDM v1.4] Add Extendable::getExtensionByName (#210)

sebalaig:

#210 

**Other information**:


include/powsybl/iidm/Extendable.hxx
```diff
@@ -55,6 +55,22 @@ E& Extendable::getExtension() {
     return const_cast<E&>(static_cast<const Extendable*>(this)->getExtension<E>());
 }
 
+template <typename E, typename>
+stdcxx::CReference<E> Extendable::getExtensionByName(const std::string& name) const {
+    auto it = m_extensionsByName.find(name);
+    if (it != m_extensionsByName.end()) {
+        if (stdcxx::isInstanceOf<E>(*it->second)) {
```
sebalaig, line 62: Sonars says to merge both conditions

mathbagu: I agree with Sonar

test/iidm/ExtensionTest.cpp
```diff
@@ -0,0 +1,125 @@
...
+    const Load& cLoad = load;
+
+    BOOST_CHECK(!cLoad.getExtensionByName<LoadExtension>("test"));
+    BOOST_CHECK(!load.getExtensionByName<LoadExtension>("test"));
+
+    BOOST_CHECK(!cLoad.getExtensionByName<BatteryExtension>("test"));
+    BOOST_CHECK(!load.getExtensionByName<BatteryExtension>("test"));
+
+    auto cExt = cLoad.getExtensionByName<LoadExtension>("loadExtension");
+    auto ext = load.getExtensionByName<LoadExtension>("loadExtension");
```
mathbagu, line 114: I would use LoadFooExt and LoadBarExt that have been created for unit testing. This would reduce the amount of code needed to test the new feature. What do you think?

mathbagu (changes requested):

test/iidm/ExtensionTest.cpp
sebalaig: Agree! Thanks for the tip! I thought about using LoadMockExt and LoadQuxExt which are in tests but they were in submodule `iidm::converter::xml::extension`, so using them from iidm was not very conventional...

mathbagu (approved):

Issue #71, opened 2020-03-27
Implement SimpleAnonymizer


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
The only anonymizer we have is the FakeAnonymizer. We should consider implementing the SimpleAnonymizer.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**

* **What is the motivation / use case for changing the behavior?**
Have both Java and C++ implementations as close as possible.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

sebalaig:
Closed by #271

PR #271, opened 2021-01-29
[IIDM v1.4] Add SimpleAnonymizer (#71)

sebalaig:

#71

This PR adds `SimpleAnonymizer` class. By using this new class and really anonymizing a network, 2 bugs were found in anonymization/deanonymization of network. Both are fixed in this PR by 2 separated commits : 
- 7761f81161f9a24299a46ecb7c449bf7dacb3fa3
- 6fb642f19c7d18727a473d9b99d9a1dbf56cbc46

**Other information**:

### DIFF
--- a/include/powsybl/iidm/converter/FakeAnonymizer.hpp
+++ b/include/powsybl/iidm/converter/FakeAnonymizer.hpp
@@ -34,6 +34,14 @@ public:
     FakeAnonymizer() = default;
 
     ~FakeAnonymizer() override = default;
+
+    FakeAnonymizer(const FakeAnonymizer&) = default;
+
+    FakeAnonymizer(FakeAnonymizer&&) noexcept = default;
+
+    FakeAnonymizer& operator=(const FakeAnonymizer&) = default;
+
+    FakeAnonymizer& operator=(FakeAnonymizer&&) noexcept = default;
 };
 
 }  // namespace converter
--- a/include/powsybl/iidm/converter/FakeAnonymizer.hpp
+++ b/include/powsybl/iidm/converter/FakeAnonymizer.hpp
@@ -17,11 +17,7 @@ namespace iidm {
 namespace converter {
 
 class FakeAnonymizer : public Anonymizer {
-public:
-    FakeAnonymizer() = default;
-
-    ~FakeAnonymizer() override = default;
-
+public:  // Anonymizer
     Country anonymizeCountry(const Country& country) override;
 
     std::string anonymizeString(const std::string& str) override;
@@ -29,6 +25,15 @@ public:
     Country deanonymizeCountry(const Country& country) const override;
 
     std::string deanonymizeString(const std::string& anonymousStr) const override;
+
+    void read(std::istream& stream) override;
+
+    void write(std::ostream& stream) const override;
+
+public:
+    FakeAnonymizer() = default;
+
+    ~FakeAnonymizer() override = default;
 };
 
 }  // namespace converter
--- a/include/powsybl/iidm/converter/SimpleAnonymizer.hpp
+++ b/include/powsybl/iidm/converter/SimpleAnonymizer.hpp
@@ -33,6 +33,9 @@ public:  // Anonymizer
     void write(std::ostream& stream) const override;
 
 public:
+    static std::string getAlpha(unsigned long num);
+
+public:
     SimpleAnonymizer() = default;
 
     ~SimpleAnonymizer() override = default;
@@ -46,9 +49,6 @@ public:
     SimpleAnonymizer& operator=(SimpleAnonymizer&&) = default;
 
     unsigned long getStringCount() const;
-
-private:
-    static std::string getAlpha(unsigned long num);
 
 private:
     using Mapping = boost::bimap<std::string, std::string>;
--- a/include/powsybl/iidm/converter/SimpleAnonymizer.hpp
+++ b/include/powsybl/iidm/converter/SimpleAnonymizer.hpp
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#ifndef POWSYBL_IIDM_CONVERTER_SIMPLEANONYMIZER_HPP
+#define POWSYBL_IIDM_CONVERTER_SIMPLEANONYMIZER_HPP
+
+#include <map>
+
+#include <powsybl/iidm/converter/Anonymizer.hpp>
+
+namespace powsybl {
+
+namespace iidm {
+
+namespace converter {
+
+class SimpleAnonymizer : public Anonymizer {
+public:  // Anonymizer
+    Country anonymizeCountry(const Country& country) override;
+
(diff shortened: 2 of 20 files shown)


src/iidm/converter/SimpleAnonymizer.cpp
```diff
@@ -0,0 +1,92 @@
...
+Country SimpleAnonymizer::anonymizeCountry(const Country& country) {
+    return country;
+}
+
+std::string SimpleAnonymizer::anonymizeString(const std::string& str) {
+    auto it = m_mapping.find(str);
+    if (it != m_mapping.end()) {
+        return it->second;
+    }
+    std::pair<std::map<std::string, std::string>::iterator, bool> status = m_mapping.emplace(str, getAlpha(m_mapping.size() + 1));
```
sebalaig, line 33: Use auto

sebalaig (commented):
Also add option `--mapping-file` to iidm-benchmark program

include/powsybl/iidm/converter/FakeAnonymizer.hpp
```diff
@@ -17,18 +17,23 @@ namespace iidm {
...
     std::string deanonymizeString(const std::string& anonymousStr) const override;
+
+    void read(std::istream& stream) override;
+
+    void write(std::ostream& stream) const override;
+
+public:
+    FakeAnonymizer() = default;
+
+    ~FakeAnonymizer() override = default;
```
mathbagu, line 36: Add all constructors and assignment operators

include/powsybl/iidm/converter/SimpleAnonymizer.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include <powsybl/iidm/converter/Anonymizer.hpp>
+
+namespace powsybl {
+
+namespace iidm {
+
+namespace converter {
+
+class SimpleAnonymizer : public Anonymizer {
```
mathbagu, line 21: To be discussed: maybe we should introduce the StringAnonymizer class in the util package

include/powsybl/iidm/converter/SimpleAnonymizer.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+    unsigned long getStringCount() const;
+
+private:
+    static std::string getAlpha(unsigned long num);
+
+private:
+    std::map<std::string, std::string> m_mapping;
+
+    std::map<std::string, std::string> m_reverseMapping;
```
mathbagu, line 58: Consider Boost::Bimap

src/iidm/converter/SimpleAnonymizer.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+std::string SimpleAnonymizer::anonymizeString(const std::string& str) {
+    auto it = m_mapping.find(str);
+    if (it != m_mapping.end()) {
+        return it->second;
+    }
+    auto status = m_mapping.emplace(str, getAlpha(m_mapping.size() + 1));
+    m_reverseMapping[status.first->second] = str;
+    return status.first->second;
+}
```
mathbagu, line 36: I'm not sure how empty strings are handled, but in the java version it's a special case.

src/iidm/converter/xml/NetworkXml.cpp
```diff
@@ -267,9 +268,12 @@ std::unique_ptr<Anonymizer> NetworkXml::write(std::ostream& ostream, const Netwo
...
 
-    // FIXME(sebalaig): for now on, only one kind of anonymizer = FakeAnonymizer
-    // later, a real anonymizer will be instantiated depending on options.isAnonymized()
-    std::unique_ptr<Anonymizer> anonymizer(stdcxx::make_unique<FakeAnonymizer>());
+    std::unique_ptr<Anonymizer> anonymizer;
+    if (options.isAnonymized()) {
+        anonymizer = stdcxx::make_unique<SimpleAnonymizer>();
+    } else {
+        anonymizer = stdcxx::make_unique<FakeAnonymizer>();
+    }
```
mathbagu, line 276: Use a ternary operator? If you have an issue with the return type, use `make_unique<Anonymizer, SimpleAnonymizer>` and `make_unique<Anonymizer, FakeAnonymizer>`

tools/benchmark/Benchmark.cpp
```diff
@@ -123,8 +126,35 @@ int main(int argc, char** argv) {
...
+        if (exportOptions.isAnonymized()) {
+            const std::string& outputMappingFile = boost::filesystem::path(outputFile).replace_extension("csv").string();
+            std::ofstream outputMappingStream(outputMappingFile);
+            if (!outputMappingStream.is_open()) {
+                std::cerr << stdcxx::format("Unable to open mapping file '%1%' for writing", outputMappingFile) << std::endl;
+                return EXIT_FAILURE;
+            }
+            anonymizer->write(outputMappingStream);
+            outputMappingStream.close();
+        }
```
mathbagu, line 157: I think we miss something regarding the Java implementation. The XMLImporter looks for a mapping file near the XIIDM file at the import.
In my mind, the `MAPPING_FILE` was only used for the export, but as the name is based on the output-file, maybe we have done something complicated for nothing.

mathbagu (changes requested):

include/powsybl/iidm/converter/SimpleAnonymizer.hpp
sebalaig: Discussed : as `StringAnonymizer ` would be used by `SimpleAnonymizer` for now on, its would not be relevant. So it would be refactored later if needed.

src/iidm/converter/SimpleAnonymizer.cpp
sebalaig: You're right, I forgot a special case to avoid adding an entry in the map...

src/iidm/converter/xml/NetworkXml.cpp
```diff
@@ -217,11 +220,21 @@ Network NetworkXml::read(const std::string& /*filename*/, std::istream& is, cons
...
+        if (!csvStream.is_open()) {
+            throw PowsyblException(stdcxx::format("Unable to open file '%1%' for reading", csvPath.filename().string()));
+        }
+        anonymizer->read(csvStream);
+    } else {
+        anonymizer = stdcxx::make_unique<FakeAnonymizer>();
+    }
 
-    NetworkXmlReaderContext context(anonymizer, reader, options, version);
+    NetworkXmlReaderContext context(*anonymizer, reader, options, version);
```
mathbagu, line 237: To be discussed: I'm not sure this is correct to get the pointer like this. How is it converted to a reference? As the anonymizer is now always created in the implementation and not exposed in the API, should we move the instance to make context the new owner. I think the lifetime of the context and the anonymizer are the same

src/iidm/converter/xml/NetworkXml.cpp
```diff
@@ -321,6 +332,15 @@ std::unique_ptr<Anonymizer> NetworkXml::write(const std::string& /*filename*/, s
     writer.writeEndElement();
     writer.writeEndDocument();
 
+    if (options.isAnonymized()) {
+        boost::filesystem::path csvPath = boost::filesystem::path(filename).replace_extension("csv");
+        boost::filesystem::ofstream oStream(csvPath);
```
mathbagu, line 337: To be consistent, rename `oStream` to `csvStream`?

mathbagu (approved):

PR #273, opened 2021-02-11
Refactoring XML read/write of a Network API

sebalaig:

No

XML serialization API update.

Update XML read/write API to prepare implementation of real anonymization (See PR #271) :
Functions available from public API of `Network` enabling to read/write do not pass filename which is needed so that CSV file can be guessed from input file like java Importers does.

Filename is always given when calling `Network::readXml` and `Network::writeXml` so that deanonymization can be done correctly.

⚠️ This is a really important API breaking change : 
 - the existing `Network::readXml(const std::string& networkStr)` has been removed
 - a new version `readXml(const boost::filesystem::path& path, const converter::ImportOptions& options = converter::ImportOptions())` is **silently called instead** because `boost::filesystem::path` constructor is not explicit 

- [x] The *Breaking Change* or *Deprecated* label has been added

**Other information**:

### DIFF
--- a/src/iidm/Network.cpp
+++ b/src/iidm/Network.cpp
@@ -10,6 +10,7 @@
 #include <sstream>
 #include <unordered_set>
 
+#include <boost/filesystem/fstream.hpp>
 #include <boost/range/join.hpp>
 
 #include <powsybl/iidm/Battery.hpp>
@@ -44,26 +45,28 @@ namespace powsybl {
 
 namespace iidm {
 
-Network Network::readXml(const std::string& data) {
-    std::stringstream stream(data);
-    return readXml(stream);
+Network Network::readXml(const boost::filesystem::path& path, const converter::ImportOptions& options) {
+    boost::filesystem::ifstream is(path);
+    if (!is.is_open()) {
+        throw PowsyblException(stdcxx::format("Unable to open file '%1%' for reading", path.filename().string()));
+    }
+    return readXml(path.filename().string(), is, options);
 }
 
-Network Network::readXml(std::istream& istream) {
-    return readXml(istream, converter::ImportOptions(), converter::FakeAnonymizer());
+Network Network::readXml(const std::string& filename, std::istream& istream, const converter::ImportOptions& options) {
+    return converter::xml::NetworkXml::read(filename, istream, options);
 }
 
-Network Network::readXml(std::istream& istream, const converter::ImportOptions& options, const converter::Anonymizer& anonymizer) {
-    return converter::xml::NetworkXml::read(istream, options, anonymizer);
+void Network::writeXml(const boost::filesystem::path& path, const Network& network, const converter::ExportOptions& options) {
+    boost::filesystem::ofstream os(path);
+    if (!os.is_open()) {
+        throw PowsyblException(stdcxx::format("Unable to open file '%1%' for writing", path.filename().string()));
+    }
+    writeXml(path.filename().string(), os, network, options);
 }
 
-std::unique_ptr<converter::Anonymizer> Network::writeXml(std::ostream& ostream, const Network& network) {
-    converter::ExportOptions options;
-    return writeXml(ostream, network, options);
-}
-
-std::unique_ptr<converter::Anonymizer> Network::writeXml(std::ostream& ostream, const Network& network, const converter::ExportOptions& options) {
-    return converter::xml::NetworkXml::write(ostream, network, options);
+void Network::writeXml(const std::string& filename, std::ostream& ostream, const Network& network, const converter::ExportOptions& options) {
+    converter::xml::NetworkXml::write(filename, ostream, network, options);
 }
 
 Network::Network(const std::string& id, const std::string& sourceFormat) :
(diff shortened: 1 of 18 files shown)


src/iidm/Network.cpp
```diff
@@ -44,26 +45,28 @@ namespace powsybl {
...
-}
-
-Network Network::readXml(std::istream& istream) {
-    return readXml(istream, converter::ImportOptions(), converter::FakeAnonymizer());
+Network Network::readXml(const boost::filesystem::path& path, const converter::ImportOptions& options) {
+    boost::filesystem::ifstream is(path);
+    if (!is.is_open()) {
+        throw PowsyblException(stdcxx::format("Unable to open file '%1%' for reading", path.filename().string()));
+    }
+    return readXml(path.filename().string(), is, options);
```
mathbagu, line 53: I would call NetworkXml directly to avoid one useless method call

src/iidm/Network.cpp
```diff
@@ -44,26 +45,28 @@ namespace powsybl {
...
 
-std::unique_ptr<converter::Anonymizer> Network::writeXml(std::ostream& ostream, const Network& network) {
-    converter::ExportOptions options;
-    return writeXml(ostream, network, options);
+void Network::writeXml(const boost::filesystem::path& path, const Network& network, const converter::ExportOptions& options) {
+    boost::filesystem::ofstream os(path);
+    if (!os.is_open()) {
+        throw PowsyblException(stdcxx::format("Unable to open file '%1%' for writing", path.filename().string()));
+    }
+    writeXml(path.filename().string(), os, network, options);
```
mathbagu, line 65: Same remark here

test/iidm/converter/xml/NetworkXmlTest.cpp
```diff
@@ -110,21 +114,20 @@ BOOST_AUTO_TEST_CASE(FromParameters) {
     stdcxx::Properties properties;
     properties.set(ImportOptions::THROW_EXCEPTION_IF_EXTENSION_NOT_FOUND, "true");
 
-    FakeAnonymizer anonymizer;
-    const Network& network = Network::readXml(stream, ImportOptions(properties), anonymizer);
+    const Network& network = Network::readXml("network.xiidm", stream, ImportOptions(properties));
 
     std::stringstream ostream;
-    Network::writeXml(ostream, network, ExportOptions(properties));
+    Network::writeXml(stdcxx::format("%1%.xiidm", network.getId()), ostream, network, ExportOptions(properties));
```
mathbagu, line 120: Use a local variable?

test/iidm/converter/xml/IdentifiableExtensionXmlSerializerTest.cpp
```diff
@@ -56,7 +56,8 @@ BOOST_AUTO_TEST_CASE(TerminalExtension) {
     Load& load = network.getLoad("LOAD");
     load.addExtension(stdcxx::make_unique<extensions::TerminalMockExt>(load));
 
-    Network network2 = Network::readXml(test::converter::RoundTrip::getVersionedNetwork("eurostag-tutorial-example1-with-terminalMock-ext.xml", IidmXmlVersion::CURRENT_IIDM_XML_VERSION()));
+    std::istringstream stream(test::converter::RoundTrip::getVersionedNetwork("eurostag-tutorial-example1-with-terminalMock-ext.xml", IidmXmlVersion::CURRENT_IIDM_XML_VERSION()));
+    Network network2 = Network::readXml("network.xiidm", stream);
```
mathbagu, line 60: It's a detail, but you should store the filename in a variable and pass the same value to `getVersionedNetwork` and `readXml` methods

mathbagu (changes requested):

test/iidm/converter/xml/IdentifiableExtensionXmlSerializerTest.cpp
sebalaig: Done.

src/test/converter/RoundTrip.cpp
```diff
@@ -65,19 +67,20 @@ void RoundTrip::roundTripVersionedXmlFromMinToCurrentVersionTest(const std::stri
...
+        iidm::Network::writeXml(filename, stream, n, options);
     };
 
-    const auto& reader = [](const std::string& xml) {
-        std::istringstream stream(xml);
-        return iidm::Network::readXml(stream);
+    const auto& reader = [&filename](const std::string& networkStr) {
+        std::istringstream stream(networkStr);
+        return iidm::Network::readXml(filename, stream);
     };
```
mathbagu, line 79: I think networkStr is a bad variable name: xml seems better to me if it really contains the XML content

src/test/converter/RoundTrip.cpp
```diff
@@ -96,12 +99,13 @@ iidm::Network RoundTrip::run(const iidm::Network& network, const Writer& out, co
...
-        iidm::Network::writeXml(stream, n);
+        const std::string& filename = stdcxx::format("%1%.xiidm", n.getId());
+        iidm::Network::writeXml(filename, stream, n);
     };
 
-    const auto& reader = [](const std::string& xml) {
-        std::istringstream stream(xml);
-        return iidm::Network::readXml(stream);
+    const auto& reader = [](const std::string& networkStr) {
+        std::istringstream stream(networkStr);
```
mathbagu, line 107: Same remark

mathbagu (approved):

PR #278, opened 2021-02-24
Fix CI (boost not avaible on github actions)

sebalaig:

No

Boost is not available on CI environments provided by github actions (https://github.com/actions/virtual-environments/issues/2667)

CI is failing because boost is required and not available on environments

Boost is installed : 
 - From apt-get repository on ubuntu
 - From [official pre-built binaries](https://sourceforge.net/projects/boost/files/boost-binaries/) on windows

**Other information**:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,8 +67,9 @@ jobs:
           cscript configure.js compiler=msvc prefix=C:\thirdparties\libxml2-2.9.10 iconv=no
           nmake -f Makefile.msvc
           nmake -f Makefile.msvc install
-          wget -nv -O boost_1_72_0-msvc-14.2-64.exe https://sourceforge.net/projects/boost/files/boost-binaries/1.72.0/boost_1_72_0-msvc-14.2-64.exe/download
-          boost_1_72_0-msvc-14.2-64.exe /dir="C:\thirdparties\boost-1.72.0" /sp- /verysilent /suppressmsgboxes /norestart
+          wget -nv -O boost-installer.exe https://sourceforge.net/projects/boost/files/boost-binaries/1.72.0/boost_1_72_0-msvc-14.2-64.exe/download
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -65,6 +67,8 @@ jobs:
           cscript configure.js compiler=msvc prefix=C:\thirdparties\libxml2-2.9.10 iconv=no
           nmake -f Makefile.msvc
           nmake -f Makefile.msvc install
+          wget -nv -O boost_1_72_0-msvc-14.2-64.exe https://sourceforge.net/projects/boost/files/boost-binaries/1.72.0/boost_1_72_0-msvc-14.2-64.exe/download
+          boost_1_72_0-msvc-14.2-64.exe /dir="C:\thirdparties\boost-1.72.0" /sp- /verysilent /suppressmsgboxes /norestart
```
jsoref, line 71: You might consider having this stage write the boost path to a variable in `$GITHUB_ENV` and then just using that variable in later steps:
https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#environment-files

This way future upgrades will not have to change the lines below:

sebalaig: Done. Thanks for the suggestion.

mathbagu (approved):

.github/workflows/clang-tidy.yml
```diff
@@ -9,9 +9,12 @@ jobs:
     name: clang-tidy
     runs-on: ubuntu-latest
     steps:
+      - name: Install Boost
+        run: sudo apt install -y libboost-all-dev
+
       - name: Install clang-tidy
         run: |
-          sudo apt-get install -y clang-tidy-9
+          sudo apt install -y clang-tidy-9
```
jsoref, line 17: At least last I checked, `apt-get` (and `apt-cache`) was (were) recommended for CI because `apt` didn't have a committed non-interactive mode, and it will spit out a warning when run in such conditions.

Issue #230, opened 2020-12-11
Iso API with Java implementation


sebalaig:
* **Do you want to request a *feature* or report a *bug*?**
Documentation / API checking :
In Java (https://github.com/powsybl/powsybl-core), there is public interface, and there are several implementations.
In C++ there is only one implementation with public and private methods.

Known bug:
- Remove Leg::toString() which is not in the java API.

We have to check that public features in C++ are strictly equivalent to Java ones.

* **What is the current behavior?**
NA

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
NA

* **What is the expected behavior?**
NA

* **What is the motivation / use case for changing the behavior?**
The aim of this issue is to check the public API available in both library versions to ensure that there is no more feature available in one or the other one.

This study may highlight some potential integrity issues that would have to be discussed and/or resolved in a near future.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

Issue #223, opened 2020-11-26
Make all enums easily printable


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Refactoring

* **What is the current behavior?**
Most enums in IIDM are printable using the `operator<<` defined in Enum.hpp. It uses the `getNames()` template method that allow us to also serialize enums using `Enum::toString` and `Enum::fromString`.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Sometimes, we have to declare another `operator<<` (ExportOptions, ConnectablePosition::Direction) but we don't really understand when we have to do what.
If the operator is needed in unit tests (BOOST_CHECK_EQUAL), the operator has to be either in the `std` namespace or in the current namespace.

* **What is the expected behavior?**
Find a design that we can apply everywhere in powsybl.

* **What is the motivation / use case for changing the behavior?**
Make the implementation of enums simpler by a common design. 

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #284, opened 2021-03-02
Iso APIs: backport missing methods

sebalaig:

#230 only backport **missing and relevant** methods
`VoltageLevel::traverse` will be added in a separated PR.

**Other information**:

### DIFF
--- a/include/powsybl/iidm/Network.hpp
+++ b/include/powsybl/iidm/Network.hpp
@@ -161,6 +161,8 @@ public:
 
     ConnectedComponentsManager& getConnectedComponentsManager();
 
+    std::set<Country> getCountries() const;
+
     unsigned long getCountryCount() const;
 
     const DanglingLine& getDanglingLine(const std::string& id) const;
--- a/src/iidm/Bus.cpp
+++ b/src/iidm/Bus.cpp
@@ -141,8 +141,6 @@ double Bus::getQ() const {
         const Connectable& connectable = terminal.getConnectable();
         switch (connectable.getType()) {
             case ConnectableType::BUSBAR_SECTION:
-            case ConnectableType::SHUNT_COMPENSATOR:
-            case ConnectableType::STATIC_VAR_COMPENSATOR:
             case ConnectableType::LINE:
             case ConnectableType::TWO_WINDINGS_TRANSFORMER:
             case ConnectableType::THREE_WINDINGS_TRANSFORMER:
@@ -152,6 +150,8 @@ double Bus::getQ() const {
             case ConnectableType::GENERATOR:
             case ConnectableType::BATTERY:
             case ConnectableType::LOAD:
+            case ConnectableType::SHUNT_COMPENSATOR:
+            case ConnectableType::STATIC_VAR_COMPENSATOR:
             case ConnectableType::HVDC_CONVERTER_STATION:
                 if (!std::isnan(terminal.getQ())) {
                     q += terminal.getQ();
--- a/src/iidm/Bus.cpp
+++ b/src/iidm/Bus.cpp
@@ -126,7 +126,7 @@ double Bus::getP() const {
                 }
                 break;
             default:
-                throw new AssertionError(stdcxx::format("Unexpected connectable type: %1%", connectable.getType()));
+                throw AssertionError(stdcxx::format("Unexpected connectable type: %1%", connectable.getType()));
         }
     }
     return p;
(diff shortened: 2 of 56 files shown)


include/powsybl/iidm/Network.hpp
```diff
@@ -161,6 +161,8 @@ class Network : public Container, public VariantManagerHolder {
 
     ConnectedComponentsManager& getConnectedComponentsManager();
 
+    std::set<Country> getCountries() const;
```
mathbagu, line 164: I wonder if we should use a range here

src/iidm/Bus.cpp
```diff
@@ -100,6 +100,70 @@ Network& Bus::getNetwork() {
...
+double Bus::getQ() const {
+    if (getConnectedTerminalCount() == 0) {
+        return stdcxx::nan();
+    }
+    double q = 0;
+    for (const Terminal& terminal : getConnectedTerminals()) {
+        const Connectable& connectable = terminal.getConnectable();
+        switch (connectable.getType()) {
+            case ConnectableType::BUSBAR_SECTION:
+            case ConnectableType::SHUNT_COMPENSATOR:
```
mathbagu, line 144: This is not correct: for shunts, you have to consider the reactive power

src/iidm/Network.cpp
```diff
@@ -206,16 +206,19 @@ ConnectedComponentsManager& Network::getConnectedComponentsManager() {
     return m_variants.get().getConnectedComponentsManager();
 }
 
-unsigned long Network::getCountryCount() const {
-    std::unordered_set<Country, stdcxx::hash<Country>> countries;
+std::set<Country> Network::getCountries() const {
+    std::set<Country> countries;
```
mathbagu, line 210: Why do you change the type of the set?

test/iidm/NetworkTest.cpp
```diff
@@ -372,12 +373,20 @@ BOOST_AUTO_TEST_CASE(fictitious) {
...
+}
+
 BOOST_AUTO_TEST_CASE(country) {
     Network network = createTestNetwork();
 
     BOOST_CHECK_EQUAL(4, network.getSubstationCount());
     BOOST_CHECK_EQUAL(2, network.getCountryCount());
 
+    const std::set<Country>& countries = { Country::ES, Country::FR };
+    BOOST_CHECK(countries ==  network.getCountries());
```
mathbagu, line 388: There is an extra space. Is it possible to use BOOST_CHECK_EQUAL?

mathbagu (changes requested):

include/powsybl/iidm/Network.hpp
sebalaig: Agree, write such a code is better:
```
stdcxx::const_range<Country> Network::getCountries() const {
    const auto& filter = [](const Substation& substation) {
        return static_cast<bool>(substation.getCountry());
    };

    const auto& mapper = [](const Substation& substation) {
        return *substation.getCountry();
    };

    return getSubstations() | boost::adaptors::filtered(filter) | boost::adaptors::transformed(mapper);
}
```
But the problem is that we also have to ensure unicity, and 
```
    return boost::range::unique(getSubstations() | boost::adaptors::filtered(filter) | boost::adaptors::transformed(mapper));
```
does not compile because of a strange non assignable error (probably because it's a range of references):
```
[  0%] Building CXX object src/CMakeFiles/iidm.dir/iidm/Network.cpp.o
/usr/include/boost/optional/optional.hpp:697:58: error: object of type '(lambda at /mnt/e/Projet_SRO/rte/powsybl-iidm4cpp/src/iidm/Network.cpp:211:26)' cannot be assigned because its copy assignment operator is implicitly deleted [clang-diagnostic-error]
    void assign_value ( argument_type val ) { get_impl() = val; }
```

Both statements:
```
    return getSubstations() | boost::adaptors::filtered(filter) | boost::adaptors::transformed(mapper) | boost::adaptors::uniqued;
    return boost::adaptors::unique(getSubstations() | boost::adaptors::filtered(filter) | boost::adaptors::transformed(mapper));
```
compile but does nothing 😢

test/iidm/NetworkTest.cpp
sebalaig: >Is it possible to use BOOST_CHECK_EQUAL?

No it isn't (problem described in #223)

mathbagu (approved):

PR #299, opened 2021-03-19
Add options to handle static and static runtime builds.

gautierbureau:
Signed-off-by: Gautier Bureau <gautier.bureau@gmail.com>

- [] Docs have been added / updated (for bug fixes / features)

Enables to build only static/shared versions of the library and use static runtime on windows

Static and shared versions are always built

Choose a specific version static/shared or both. Use a static boost library. Use static runtime on windows.


CMakeLists.txt
```diff
@@ -36,9 +61,28 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_CXX_COMPILER_ID}" STRE
     # set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -D_GLIBCXX_DEBUG")
 elseif (MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4244 /wd4250 /wd4251 /wd4267 /wd4275")
+    if (NOT BUILD_SHARED_LIBS AND BUILD_STATIC_LIBS)
+        if (MSVC_STATIC_RUNTIME_LIBRARY)
+            set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
```
mathbagu, line 66: This variable has been added in cmake v3.15 (https://cmake.org/cmake/help/git-stage/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html). Is there an equivalent for older version?

CMakeLists.txt
```diff
@@ -36,9 +61,28 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_CXX_COMPILER_ID}" STRE
...
+    set(Boost_USE_STATIC_LIBS ON)
+    if (MSVC_STATIC_RUNTIME_LIBRARY)
+        set(Boost_USE_STATIC_RUNTIME ON)
+    else ()
+        set(Boost_USE_STATIC_RUNTIME OFF)
+    endif ()
+else ()
+    set(Boost_USE_STATIC_LIBS OFF)
+    set(Boost_USE_STATIC_RUNTIME OFF)
+endif ()
```
mathbagu, line 85: I propose to set `Boost_USE_STATIC_RUNTIME` according to the value of `MSVC_STATIC_RUNTIME_LIBRARY` even if `BOOST_STATIC_LIBS` is not set.
```
set(Boost_USE_STATIC_LIBS ${BOOST_STATIC_LIBS})
set(Boost_USE_STATIC_RUNTIME ${MSVC_STATIC_RUNTIME_LIBRARY})
```

CMakeLists.txt
```diff
@@ -24,6 +30,25 @@ option(BUILD_DOXYGEN "Enable/Disable the generation of the Doxygen." OFF)
 option(BUILD_EXAMPLES "Enable/Disable the compilation of the examples." OFF)
 option(BUILD_TESTS "Enable/Disable the compilation of unit tests." ON)
 option(BUILD_TOOLS "Enable/Disable the compilation of tools." ON)
+option(BUILD_SHARED_LIBS "Enable/disable build of shared libraries." ON)
+option(BUILD_SHARED_EXTENSIONS "Enable/disable build of shared extensions." ON)
+option(BUILD_STATIC_LIBS "Enable/disable build of static libraries." ON)
+option(BUILD_STATIC_EXTENSIONS "Enable/disable build of static extensions." ON)
```
mathbagu, line 36: It's arguable, but I think we should not allow people to mix static and dynamic libraries. I propose to keep only `BUILD_SHARED_LIBS` and `BUILD_STATIC_LIBS`. I wonder if we should add an option for each extension library

CMakeLists.txt
```diff
@@ -24,6 +30,25 @@ option(BUILD_DOXYGEN "Enable/Disable the generation of the Doxygen." OFF)
...
+option(MSVC_STATIC_RUNTIME_LIBRARY "Enable/disable static runtime on Windows." OFF)
+option(BOOST_STATIC_LIBS "Use static version of boost libraries." OFF)
+
+if (BOOST_STATIC_LIBS AND BUILD_SHARED_LIBS)
+    set(BOOST_STATIC_LIBS OFF)
+    message(WARNING "BOOST_STATIC_LIBS disabled if you want to build shared libraries. Use this option with BUILD_SHARED_LIBS=OFF")
+endif ()
```
mathbagu, line 43: I would raise an error in that case instead of making an arbitrary choice.

CMakeLists.txt
```diff
@@ -24,6 +30,25 @@ option(BUILD_DOXYGEN "Enable/Disable the generation of the Doxygen." OFF)
...
+
+if (BUILD_TESTS AND NOT BUILD_SHARED_LIBS)
+    message(FATAL_ERROR "BUILD_TESTS is ON but BUILD_SHARED_LIBS is OFF, you should turn ON BUILD_SHARED_LIBS or turn off BUILD_TESTS.")
+endif ()
```
mathbagu, line 47: Maybe we have to do something similar for the examples. Maybe we can find a way to link the examples and the tests to the IIDM library, using a postfix for example (empty or -static).

CMakeLists.txt
gautierbureau: done

CMakeLists.txt
gautierbureau: done

gautierbureau:
Following a discussion with @mathbagu this PR will be put on hold indefinitely. Some conclusions on the proposed solution and future work that could be bone:
- as it is now the unit tests can now be launched with a static or shared build but with separate cmake build directories. Previously unit tests was only tested against the shared versions of the library.
- One issue with the proposed solutions is that the methods `ExtensionProviders::loadExtensions` and `ExtensionProviders::loadLibrary` are not tested anymore because the class `ExtensionFixture` is not used anymore to load extensions but rather the `registerExtension`method is used.
- Following the two previous issues we stated that unit tests should be launched in static and shared versions in the same cmake build and that there should be distinct target to test them to have code compiled in one version of the target and not the other. For exemple for the shared unit tests  `ExtensionFixture` should still be used and for the the static version a new class similar to what it is proposed here should be used to `registerExtension`.

Issue #302, opened 2021-03-19
Put registerExtension in public


gautierbureau:
* **Do you want to request a *feature* or report a *bug*?**
registerExtension should be public to be able to register extensions in a application where extensions are used and linked directly to the main binary.

* **What is the current behavior?**
Only way to load extensions is with loadExtensions.

* **What is the expected behavior?**
Call registerExtension from client code
`registerExtension(stdcxx::make_unique<ExtensionClass>())`

* **What is the motivation / use case for changing the behavior?**
When using extensions in a main program if the methods of the extensions are used the user is force to link the .so of the extension (or .a) then symbols are already loaded and no need for a dlopen to be called.

mathbagu:
Fixed in #303

PR #303, opened 2021-03-19
Make registerExtension public

mathbagu:
- [X] Docs have been added / updated (for bug fixes / features)

#302 

Feature

When we try to use the extensions, we have to link with the corresponding extension library. Currently, there is no way to register manually an extension, so we have to load the same library at runtime to be able to serialize the network.

The `ExtensionProviders::registerExtension` is now public to make it possible to register an extension. This PR also provide an example to show how to use it

**Other information**:


examples/example3/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) 2020, RTE (http://www.rte-france.com)
```
sebalaig, line 2: Copyright (c) *2021*

sebalaig (approved):

PR #311, opened 2021-04-09
Fix special method declaration order

mathbagu:
- [X] Docs have been added / updated (for bug fixes / features)

N/A

Coding style

The copy and move constructors are declared after the destructor

The copy and move constructors are declared before the destructor, with other contructors

**Other information**:


sebalaig (changes requested):
It looks like the following files should also be updated:
- ExtensionAdder.hpp
- SlackTerminalAdder.hpp

Additional/related inconsistencies:
- Destructor is declared twice in SwitchAdder

mathbagu:
@sebalaig I don't understand your last comment. Do you run a script and miss that the `SwitchAdder` is declared twice, one time for each kind of view?

sebalaig (approved):
@mathbagu I added a few changes right now. It's ok for me now.

Issue #332, opened 2021-06-01
Optimise Network.getBusBreakerView.getBus(String) in case if bus/breaker topology


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the expected behavior?**
Backport of https://github.com/powsybl/powsybl-core/pull/1645

sebalaig:
Closed by PR #348

PR #348, opened 2021-08-03
Optimise getBus in Bus/Breaker view for default Network implementation (#332)

sebalaig:

#332 

**Other information**:

### DIFF
--- a/include/powsybl/iidm/Network.hpp
+++ b/include/powsybl/iidm/Network.hpp
@@ -53,6 +53,12 @@ class TwoWindingsTransformer;
 class VoltageLevel;
 class VscConverterStation;
 
+namespace bus_breaker_voltage_level {
+
+class BusBreakerViewImpl;
+
+}  // namespace bus_breaker_voltage_level
+
 class Network : public Container, public VariantManagerHolder {
 public:
     using BusBreakerView = network::BusBreakerView;
@@ -382,6 +388,8 @@ private:
     friend class VoltageLevel;
 
     friend class NodeBreakerVoltageLevel;
+
+    friend class bus_breaker_voltage_level::BusBreakerViewImpl;
 
 private:
     stdcxx::DateTime m_caseDate;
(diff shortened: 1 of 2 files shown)


include/powsybl/iidm/Network.hpp
```diff
@@ -53,6 +53,12 @@ class TwoWindingsTransformer;
 class VoltageLevel;
 class VscConverterStation;
 
+namespace bus_breaker_voltage_level {
+
+class BusBreakerViewImpl;
```
sebalaig, line 58: This is really ugly... but I could not find how to do that properly / simply...

src/iidm/BusBreakerVoltageLevelViews.cpp
```diff
@@ -30,6 +30,10 @@ stdcxx::CReference<Bus> BusBreakerViewImpl::getBus(const std::string& busId) con
 }
 
 stdcxx::Reference<Bus> BusBreakerViewImpl::getBus(const std::string& busId) {
+    const auto& bus = m_voltageLevel.getNetwork().getIndex().find<Bus>(busId);
```
mathbagu, line 33: I think you can solve your issue using `voltageLevel.getNetwork().find<Bus>(busId)`. This method is in the public API and will delegate the call to the `NetworkIndex`. You should not need to make `BusBreakerViewImpl` and `Network` friend anoymore.

Have you test it's working well for configuredBus (i.e. the bus is found in the index and not through the voltage level objects)

mathbagu (changes requested):

src/iidm/BusBreakerVoltageLevelViews.cpp
sebalaig: Oh thanks 👍 

Yes I checked it works also for ConfiguredBus by debugging it : 
image

mathbagu (approved):

Issue #333, opened 2021-06-01
MergedBus::getConnectedTerminalCount does not use the connected terminals but all


mathbagu:
* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the expected behavior?**
Backport of https://github.com/powsybl/powsybl-core/pull/1647

PR #349, opened 2021-08-03
Bug fix in MergedBus : getConnectedTerminalCount did not use the connected terminals but all (#333)

sebalaig:
⚠️ To be discussed : this issue was already OK in C++, I just added tests that where added in Java PR (https://github.com/powsybl/powsybl-core/pull/1647), but it may not be relevant...

#333 

**Other information**:

### DIFF
--- a/src/iidm/MergedBus.cpp
+++ b/src/iidm/MergedBus.cpp
@@ -11,6 +11,7 @@
 #include <cmath>
 
 #include <boost/range/adaptor/transformed.hpp>
+#include <boost/range/numeric.hpp>
 
 #include <powsybl/PowsyblException.hpp>
 #include <powsybl/iidm/Terminal.hpp>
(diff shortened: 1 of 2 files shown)


src/iidm/MergedBus.cpp
```diff
@@ -66,12 +67,11 @@ stdcxx::Reference<Component> MergedBus::getConnectedComponent() {
...
-    unsigned long count = 0;
-    for (const auto& it : m_buses) {
-        count += it.get().getConnectedTerminalCount();
-    }
+    const auto& mapper = [](const std::reference_wrapper<ConfiguredBus>& bus) {
+        return bus.get().getConnectedTerminalCount();
+    };
 
-    return count;
+    return boost::accumulate(m_buses | boost::adaptors::transformed(mapper), 0);
```
mathbagu, line 74: To be checked: is it more efficient written like this?

mathbagu (approved):

src/iidm/MergedBus.cpp
sebalaig: Well, in release mode, execution time of this function are stricly the same (0.0003 ms :/)
In debug mode, it is slower
=> I will revert

PR #354, opened 2021-08-26
Fix multiple extension add

sebalaig:
⚠️ Target branch to be checked

No

Bug fix

When 2 extensions of the same type are added to an Extendable, the extension of the given type always remain the first added.

Reproduction case:
```
    Network network("test", "test");
    network.setCaseDate(stdcxx::DateTime::parse("2016-06-27T12:27:58.535+02:00"));
    Substation& substation = network.newSubstation()
        .setId("S")
        .setCountry(Country::FR)
        .add();

    substation.newExtension<EntsoeAreaAdder>().withCode(EntsoeGeographicalCode::FR).add();
    substation.newExtension<EntsoeAreaAdder>().withCode(EntsoeGeographicalCode::BE).add();

    auto& extension = substation.getExtension<EntsoeArea>();
    BOOST_CHECK(static_cast<int>(EntsoeGeographicalCode::BE) == static_cast<int>(extension.getCode()));
```

When 2 extensions of the same type are added to an Extendable, the extension of the given type replaces the previous one.

**Other information**:

### DIFF
--- a/src/iidm/Extendable.cpp
+++ b/src/iidm/Extendable.cpp
@@ -28,6 +28,8 @@ Extendable::Extendable(Extendable&& extendable) noexcept :
 
 void Extendable::addExtension(std::unique_ptr<Extension>&& extension) {
     extension->setExtendable(*this);
+    m_extensionsByName.erase(extension->getName());
+    m_extensionsByType.erase(extension->getType());
     auto it = m_extensionsByName.emplace(std::make_pair(extension->getName(), std::move(extension)));
 
     std::reference_wrapper<Extension> refExtension = *it.first->second;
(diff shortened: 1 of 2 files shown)


src/iidm/Extendable.cpp
```diff
@@ -28,6 +28,8 @@ Extendable::Extendable(Extendable&& extendable) noexcept :
 
 void Extendable::addExtension(std::unique_ptr<Extension>&& extension) {
     extension->setExtendable(*this);
+    m_extensionsByName.erase(extension->getName());
+    m_extensionsByType.erase(extension->getType());
```
mathbagu, line 32: To be discussed: I wonder if it would be better to use the `operator[]` that will create a new entry if the key doesn't exist and replace the value otherwise.

Another option would be to create a utility function in `map.hpp` to provide such method.

PR #355, opened 2021-08-27
Make Extension constructor private (#342)

### LINKED ISSUE #342: Make extension constructor private
Bug

Currently, it's possible to create an extension directly, without using the adder.

You can then create an extension that is associated to an identifiable, but this extension is not registered in the identifiable. The extension is lost in the XML export

Forbid the usage of the constructor by making it private.

  - PowSyBl Version: ...
  - OS Version: ...

sebalaig:

Closes #342 

**Other information**:


mathbagu (commented):
If I remember well, sonar suggest to avoid using `new`. This is the reason why I introduced the `make_unique` function. Making the constructor private, you need to make `make_unique` friend to access to the constructor.
I think we will discuss about this PR on monday.

mathbagu (approved):

PR #368, opened 2021-09-23
Add a bus cache to network implementation (#335)

sebalaig:

**Other information**:

### DIFF
--- a/include/powsybl/iidm/NetworkVariant.hpp
+++ b/include/powsybl/iidm/NetworkVariant.hpp
@@ -8,6 +8,7 @@
 #ifndef POWSYBL_IIDM_NETWORKVARIANT_HPP
 #define POWSYBL_IIDM_NETWORKVARIANT_HPP
 
+#include <powsybl/iidm/BusCache.hpp>
 #include <powsybl/iidm/ConnectedComponentsManager.hpp>
 #include <powsybl/iidm/SynchronousComponentsManager.hpp>
 #include <powsybl/iidm/Variant.hpp>
@@ -20,6 +21,9 @@ namespace iidm {
 class Network;
 
 namespace network {
+
+class BusBreakerView;
+class BusView;
 
 class VariantImpl : public Variant<Network, VariantImpl> {
 public: // Variant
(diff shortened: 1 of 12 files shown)


include/powsybl/iidm/NetworkVariant.hpp
```diff
@@ -40,6 +44,10 @@ class VariantImpl : public Variant<Network, VariantImpl> {
 
     VariantImpl& operator=(VariantImpl&&) noexcept = delete;
 
+    const BusCache& getBusBreakerViewCache() const;
```
sebalaig, line 47: To be discussed: this is necessary because cache must be available from BusBreaker*Impl (so that getBus() is called on. IMO that would be ugly to add friendship for those impl classes...

include/powsybl/iidm/NetworkVariant.hpp
```diff
@@ -48,10 +56,19 @@ class VariantImpl : public Variant<Network, VariantImpl> {
 
     SynchronousComponentsManager& getSynchronousComponentsManager();
 
+private:
+    friend class network::BusView;
```
sebalaig, line 60: This is needed for `BusBreakerView` to call `invalidateCache()` which is not const...

src/iidm/BusBreakerVoltageLevelViews.cpp
```diff
@@ -130,11 +130,11 @@ BusViewImpl::BusViewImpl(BusBreakerVoltageLevel& voltageLevel) :
 }
 
 stdcxx::CReference<Bus> BusViewImpl::getBus(const std::string& busId) const {
-    return stdcxx::cref<Bus>(m_voltageLevel.getMergedBus(busId, false));
+    return m_voltageLevel.getNetwork().getVariants().get().getBusViewCache().getBus(busId);
 }
 
 stdcxx::Reference<Bus> BusViewImpl::getBus(const std::string& busId) {
-    return stdcxx::ref<Bus>(m_voltageLevel.getMergedBus(busId, false));
+    return stdcxx::ref(m_voltageLevel.getNetwork().getVariants().get().getBusViewCache().getBus(busId));
```
sebalaig, line 137: To be discussed: in Java, we did not call getMergedBus :/. Is it a previous bug in C++?

PR #369, opened 2021-10-01
SV utility class. Review and Refactoring (#356)

sebalaig:

Related to #356 

**Other information**:


src/iidm/util/LinkData.cpp
```diff
@@ -0,0 +1,42 @@
...
+                                                 double ratio2, double angle2, const std::complex<double>& ysh1, const std::complex<double>& ysh2) {
+    std::complex<double> a1 = std::polar(ratio1, angle1);
+    std::complex<double> a2 = std::polar(ratio2, angle2);
+
+    std::complex<double> ytr;
+    if (r == 0.0 && x == 0.0) {
+        ytr = std::complex<double>();
+    } else {
+        ytr = std::complex<double>(1) / std::complex<double>(r, x);
+    }
```
mathbagu, line 26: I would write this differently: from my point of view, the line 23 is useless because you assign the `ytr` variable with the same value. In java, you need to do that because by default, the value is initialized to `null`. 

In C++, you can simply negate the test in order to initialize `ytr`. What do you think?

src/iidm/util/SV.cpp
```diff
@@ -76,114 +76,218 @@ double SV::getQ() const {
...
+        s = flowAtEnd2(adm, v1, v);
+        otherSide = Branch::Side::TWO;
+    } else {
+        std::complex<double> v2 = std::polar(m_u, m_a * stdcxx::toRadians);
+        std::complex<double> s2(m_p, m_q);
+        v = voltageAtEnd1(adm, v2, s2);
+        s = flowAtEnd1(adm, v, v2);
+        otherSide = Branch::Side::ONE;
+    }
+    return {std::real(s), std::imag(s), std::abs(v), std::arg(v) * stdcxx::toDegrees, otherSide};
```
mathbagu, line 168: I'm not a really big fan of this syntax :)
If you decide to do so, why don't you use it when you pass a complex at line 120 or just below at line 172?

src/iidm/util/LinkData.cpp
sebalaig: I fully agree with you, it's useless to call default constructor twice :)

src/iidm/util/SV.cpp
sebalaig: I'm not a fan either... but it was recommended by clang-tidy : 
```
/mnt/e/Projet_SRO/rte/powsybl-iidm4cpp-master/src/iidm/util/SV.cpp:168:12: error: avoid repeating the return type from the declaration; use a braced initializer list instead [modernize-return-braced-init-list,-warnings-as-errors]
    return SV(std::real(s), std::imag(s), std::abs(v), std::arg(v) * stdcxx::toDegrees, otherSide);
```

It is not problematic on line 120 or 172, maybe because they are constexpr?

mathbagu (approved):

PR #409, opened 2021-12-17
Fix stdcxx::ref(std::unique_ptr) and stdcxx::cref(std::unique_ptr)

sebalaig:

No

Bug fix

The library crashes by writing such code:

```
std::unique_ptr<int> ptr;
auto ref = stdcxx::ref<int>(ptr);  // crash here
```

The library does not crashes, the code snippet returns an invalid `stdcxx::Reference`

**Other information**:

### DIFF
--- a/include/powsybl/stdcxx/reference.hpp
+++ b/include/powsybl/stdcxx/reference.hpp
@@ -104,7 +104,7 @@ CReference<T> cref(T& reference) {
 
 template <typename T>
 CReference<T> cref(const std::unique_ptr<T>& pointer) {
-    return CReference<T>(*pointer);
+    return static_cast<bool>(pointer) ? CReference<T>(*pointer) : CReference<T>();
 }
 
 template <typename T>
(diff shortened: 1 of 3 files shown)


include/powsybl/stdcxx/reference.hpp
```diff
@@ -129,14 +129,19 @@ Reference<T> ref(T& reference) {
...
 template <typename T, typename U, typename = typename std::enable_if<std::is_base_of<T, U>::value && !std::is_same<T, U>::value>::type>
 Reference<T> ref(U& reference) {
     return Reference<T>(dynamic_cast<T&>(reference));
 }
 
+template <typename T>
+Reference<T> ref(const Reference<T>& reference) {
+    return Reference<T>(reference);
+}
+
```
sebalaig, line 144: I had to add this one because the following code crashed : 
```
    Reference<int> myRef;

    Reference<int> refFromInvalidRef = ref<int>(myRef);
    BOOST_CHECK(!refFromInvalidRef);
    ```

Explanation: because of the operator
```
    Reference::operator T&() const {
        return get();
    }
```

the following function is called
```
template <typename T>
Reference<T> ref(T& reference) {
    return Reference<T>(reference);
}
```

mathbagu (approved):

PR #414, opened 2022-01-07
[IIDM v1.6.0] Make substations optional (#365)

### LINKED ISSUE #365: Make substations optional

Feature
Backport of:
- https://github.com/powsybl/powsybl-core/pull/1796
- https://github.com/powsybl/powsybl-core/pull/1818
- https://github.com/powsybl/powsybl-core/pull/1865

  - PowSyBl Version: ...
  - OS Version: ...

sebalaig:

Closes #365 

**Other information**:


src/iidm/ThreeWindingsTransformerAdder.cpp
```diff
@@ -71,14 +89,20 @@ ThreeWindingsTransformer& ThreeWindingsTransformerAdder::add() {
...
+    }
+    if (static_cast<bool>(m_substation)) {
+        return m_substation.get().getNetwork();
+    }
+    throw PowsyblException("Three windings transformer has no container");
 }
 
 Network& ThreeWindingsTransformerAdder::getNetwork() {
-    return m_substation.getNetwork();
+    return const_cast<Network&>(static_cast<const ThreeWindingsTransformerAdder*>(this)->getNetwork());
```
sebalaig, line 102: Notice: sonar is not happy with thoses hacks... But I think it is better than duplicating const methods content...

src/iidm/VoltageLevel.cpp
```diff
@@ -245,8 +260,10 @@ void VoltageLevel::remove() {
     // Remove the topology
     removeTopology();
 
-    // Remove this voltage level from the network
-    getSubstation().remove(*this);
+    if (getSubstation()) {
```
mathbagu, line 263: It's not necessary to call the getter here? It's useful in Java to map the return of the getter and use the fluent API of the `Optional` class

src/iidm/VoltageLevel.cpp
```diff
@@ -262,6 +279,12 @@ VoltageLevel& VoltageLevel::setLowVoltageLimit(double lowVoltageLimit) {
     return *this;
 }
 
+void VoltageLevel::setNetworkRef(Network& network) {
+    if (static_cast<bool>(m_networkRef)) {
+        m_networkRef = network;
+    }
```
mathbagu, line 285: To be discuss: I think the networkRef should be settable to null (empty reference) when the voltage is detached from the network

src/iidm/util/Substations.cpp
```diff
@@ -41,28 +41,30 @@ void checkRemovability(const Substation& substation) {
 }
 
 void checkRemovability(const Substation& substation, const Branch& branch) {
-    const Substation& s1 = branch.getTerminal1().getVoltageLevel().getSubstation();
-    const Substation& s2 = branch.getTerminal2().getVoltageLevel().getSubstation();
-    if (!stdcxx::areSame(s1, substation) || !stdcxx::areSame(s2, substation)) {
+    const auto& s1 = branch.getTerminal1().getVoltageLevel().getSubstation();
+    const auto& s2 = branch.getTerminal2().getVoltageLevel().getSubstation();
+    if ((static_cast<bool>(s1) && !stdcxx::areSame(s1.get(), substation)) || (static_cast<bool>(s2) && !stdcxx::areSame(s2.get(), substation))) {
```
mathbagu, line 46: It could be interesting to extract the test `(static_cast<bool>(s1) && !stdcxx::areSame(s1.get(), substation))` in a utility function

mathbagu (commented):
From my point of view, if we keep both `getSubstation` and `getNullableSubstation`, maybe the `getSubstation` should maybe throw an exception to avoid having an alias countrary at what we have in java.

src/iidm/VoltageLevel.cpp
sebalaig: Agree. Fixed

src/iidm/util/Substations.cpp
sebalaig: Agree. I added it in Substations.cpp file.
```
bool areDifferentSubstations(const stdcxx::CReference<Substation>& s1, const Substation& s2) {
    return static_cast<bool>(s1) && !stdcxx::areSame(s1.get(), s2);
}
```

I could hav added in memory.hpp, because this function is not intuitive enough IMHO : I would expect it returns true if the first parameter is invalid but it does not => to be discussed

include/powsybl/iidm/ThreeWindingsTransformer.hpp
```diff
@@ -49,13 +49,17 @@ class ThreeWindingsTransformer : public Connectable {
...
+
     double getRatedU0() const;
 
     Side getSide(const Terminal& terminal) const;
 
-    const Substation& getSubstation() const;
+    stdcxx::CReference<Substation> getSubstation() const;
 
-    Substation& getSubstation();
+    stdcxx::Reference<Substation> getSubstation();
```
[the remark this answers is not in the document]
sebalaig: You're right, there is not any... I wanted to be java-like, but that's not consistent with current C++ philosophy : get throw exceptions if the object does not exist. getNullableSubstation is OK.
=> updated

src/iidm/VoltageLevel.cpp
sebalaig: The bug exists in java => I posted https://github.com/powsybl/powsybl-core/issues/1960 and https://github.com/powsybl/powsybl-iidm4cpp/issues/421 to fix it later

include/powsybl/iidm/VoltageLevel.hpp
```diff
@@ -207,9 +208,18 @@ class VoltageLevel : public Container {
...
+    stdcxx::CReference<T> getConnectable(const std::string& id, const stdcxx::CReference<Substation>& substation, const stdcxx::CReference<Network>& networkRef) const;
+
     virtual void removeTopology() = 0;
 
+    void setNetworkRef(Network& network);
+
+    friend class NetworkIndex;
+
 private:
+    stdcxx::Reference<Network> m_networkRef;
```
mathbagu, line 221: Why don't we use `NetworkRef` instead like we did in Substation?

src/iidm/VoltageLevel.cpp
```diff
@@ -262,6 +271,12 @@ VoltageLevel& VoltageLevel::setLowVoltageLimit(double lowVoltageLimit) {
     return *this;
 }
 
+void VoltageLevel::setNetworkRef(Network& network) {
+    if (static_cast<bool>(m_networkRef)) {
```
mathbagu, line 275: Why this test? m_networkRef is always set except if the VoltageLevel is detached. Then, you will never enter this method. Do I miss something?

PR #434, opened 2022-05-11
Add the ability to move a Connectable (#382)

### LINKED ISSUE #382: Move connectable
Feature

Backport of https://github.com/powsybl/powsybl-core/pull/1838

  - PowSyBl Version: ...
  - OS Version: ...

sebalaig:

Closes #382 

**Other information**:


include/powsybl/iidm/Connectable.hpp
```diff
@@ -54,6 +67,18 @@ class Connectable : public Identifiable {
...
+
+    void move(Terminal& oldTerminal, const std::string& oldConnectionInfo, unsigned long node, const std::string& voltageLevelId);
+
+private:
+    void attachTerminal(Terminal& oldTerminal, const std::string& oldConnectionInfo, VoltageLevel& voltageLevel, std::unique_ptr<Terminal>&& terminal);
+
+    friend class bus_terminal::BusBreakerViewImpl;
+
+    friend class node_terminal::BusBreakerViewImpl;
+    friend class node_terminal::NodeBreakerViewImpl;
```
sebalaig, line 80: Not very proud of that... But this is necessary because `move` methods are not supposed to be in the public API...

include/powsybl/iidm/Terminal.hpp
```diff
@@ -62,6 +62,8 @@ class Terminal : public MultiVariantObject {
 
     stdcxx::Reference<Connectable> getConnectable();
 
+    virtual const std::string& getConnectionInfo() const = 0;
```
sebalaig, line 65: To be discussed: this method is uncessary in C++ as it is used in the `notifyUpdate` call in Java, which does not exist in C++.

src/iidm/NodeTerminalViews.cpp
```diff
@@ -44,6 +44,10 @@ stdcxx::Reference<Bus> BusBreakerViewImpl::getConnectableBus() {
     return voltageLevel.getCalculatedBusBreakerTopology().getConnectableBus(m_terminal.getNode());
 }
 
+void BusBreakerViewImpl::moveConnectable(const std::string& busId, bool connected) {
+    m_terminal.getConnectable().get().move(m_terminal, m_terminal.getConnectionInfo(), busId, connected);
```
sebalaig, line 48: `throw AssertionError("Not implemented");` instead ?

include/powsybl/iidm/Terminal.hpp
mathbagu: If it used only for notification, we can remove it

src/iidm/NodeTerminalViews.cpp
mathbagu: I don't understand why you suggest to throw an exception?

src/iidm/Connectable.cpp
```diff
@@ -34,6 +35,21 @@ Terminal& Connectable::addTerminal(std::unique_ptr<Terminal>&& terminal) {
...
+    voltageLevel.attach(*terminal, false);
+
+    // then we can detach the old terminal, as we now know that the new terminal is valid
+    oldTerminal.getVoltageLevel().detach(oldTerminal);
+
+    // replace the old terminal by the new terminal in the connectable
+    auto it = std::find_if(m_terminals.begin(), m_terminals.end(), [&oldTerminal](const std::unique_ptr<Terminal>& term) {
+        return stdcxx::areSame(*term, oldTerminal);
+    });
+    *it = std::move(terminal);
```
mathbagu, line 50: Is it working properly if the terminal is not found (could it really happen)?

include/powsybl/iidm/Connectable.hpp
mathbagu: I totally agree: I wonder if the methods are well placed. Looking to the C++ implementation, you can access to the attach/detach methods of the voltage level from the different views. It's probably better to move the code there to avoid these friend declarations.

Could you please ensure it is feasible or not?

mathbagu (changes requested):

src/iidm/Connectable.cpp
sebalaig: Yes, it would crash... But I don't see how it could be possible, because the `Connectable` has a list of `Terminal`, and `oldTerminal` is one of them (by definition).

Maybe we should throw an `AssertionError` would be more explicit?

include/powsybl/iidm/Connectable.hpp
sebalaig: To be discussed: I'm not sure to understand and/or I think we have an issue in C++ implementation: `attach` & `detach` methods are in the public API of the VoltageLevel in C++, and not in Java.

src/iidm/Connectable.cpp
sebalaig: TODO: write a unit test with a detached connectable

PR #463, opened 2025-07-29
Update IIDM V1.7.0

edouardChevreau:
**Please check if the PR fulfills these requirements**

- [ ] A PR or issue has been opened in all impacted repositories (if any)

**Does this PR already have an issue describing the problem?**

**What kind of change does this PR introduce?**

Update IIDM version from 1.6 to [1.7](https://github.com/powsybl/powsybl-core/wiki/IIDM-&-XIIDM-1.7-evolutions) :
  - Introduce the network validation level : 
  
    `EQUIPMENT` or `STEADY_STATE_HYPOTHESIS` values expected as `minimumValidationLevel` network's attribute :
    `<iidm:network .... minimumValidationLevel="EQUIPMENT">`

    When set to `EQUIPMENT`, following attributes are now optional. If missing in xiidm description, default values will be set and associated checks do not throw `ValidationException` thus allowing network creation at that validation level. Writing xiidm in that validation level will output those attributes anyhow (with their default values if not set).

    Newly optional attributes : 
    - p0 and q0 of a battery, load and dangling line,
    - minP, maxP, voltage regulation status and targetP of a generator,
    - active power setpoint of HVDC,
    - regulation mode of StaticVarCompensator,
    - voltage regulation status of Vsc converter station,
    - section count of shunt compensator, 
    - tap position, regulation mode and status of phase tap changer,
    - tap position, and regulating status of ratio tap changer

**What is the current behavior?**
IIDM v1.5.0

IIDM v1.6.0

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [ ] No

  New API available to manage the `ValidationLevel` of the network

**If yes, please check if the following requirements are fulfilled**

- [ ] The migration steps are described in the following section

**What changes might users need to make in their application due to this PR? (migration steps)**

**Other information**:


src/network/ScadaNetworkFactory.cpp
```diff
@@ -0,0 +1,273 @@
+/**
+ * Copyright (c) 2020, RTE (http://www.rte-france.com)
```
rosiereflo: I think we should put 2025 here, only on new files

src/network/ScadaNetworkFactory.cpp
```diff
@@ -0,0 +1,273 @@
...
+        .setNominalV(400.0)
+        .setR(1.0)
+        .setMaxP(20.0)
+        .setConverterStationId1(lcs.getId())
+        .setConverterStationId2(vcs.getId())
+        .setConvertersMode(iidm::HvdcLine::ConvertersMode::SIDE_1_INVERTER_SIDE_2_RECTIFIER)
+        .setId("hvdcline")
+        .add();
+
+    return network;
```
rosiereflo, line 268: WHy adding this here and not in test folder?

edouardChevreau: There were already a few "reusable" network factories available there. Thought this new one could come in handy to have an example of "EQUIPMENT" network ready to use.

edouardChevreau: And also we need it there to be included from within test/converter/xml for instance

